PR #36, opened 2016-09-30
View controller: Re-order misplaced methods

JohnSundell:
Some methods were accidentally placed under the `UIViewController` pragma mark, which would indicate that they are overrides from the superclass.

Those methods have now been moved to the `Private utilities` pragma mark instead.


JohnSundell:
@spotify/objc-dev

PR #37, opened 2016-09-30
HUBManager: Add convenience initializer

JohnSundell:
Since we want to support a lot of customization options for the Hub Framework (we also need them for the Spotify app), the initializer for `HUBManager` is quite large. However, most of its arguments are optional, so with this change a convenience initializer is added that only takes the required arguments.

Also update the demo app to using this new initializer.

### DIFF
--- a/sources/HUBManager.m
+++ b/sources/HUBManager.m
@@ -47,6 +47,19 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)initWithComponentLayoutManager:(id<HUBComponentLayoutManager>)componentLayoutManager
                       componentFallbackHandler:(id<HUBComponentFallbackHandler>)componentFallbackHandler
+{
+    return [self initWithComponentLayoutManager:componentLayoutManager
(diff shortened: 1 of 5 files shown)


JohnSundell:
@spotify/objc-dev

JohnSundell:
@Goos Me too 😢

sources/HUBManager.m
```diff
@@ -47,6 +47,19 @@ @implementation HUBManager
 
 - (instancetype)initWithComponentLayoutManager:(id<HUBComponentLayoutManager>)componentLayoutManager
                       componentFallbackHandler:(id<HUBComponentFallbackHandler>)componentFallbackHandler
+{
+    return [self initWithComponentLayoutManager:componentLayoutManager
+                       componentFallbackHandler:componentFallbackHandler
+                      connectivityStateResolver:nil imageLoaderFactory:nil
```
rastersize: Move `imageLoaderFactory:nil` to its own line.

JohnSundell: Ah of course, silly mistake! Will fix 👍

Issue #42, opened 2016-09-30
Add tooling to support having static and dynamic targets synced and tested


rastersize:
This is needed to make life with Carthage better since they it doesn’t support static libraries.

PR #45, opened 2016-09-30
Fix Carthage erroring out trying to build

### LINKED ISSUE #39: Carthage support is currently not functional
Hey! Just wanted to start by saying thanks for open sourcing this! Been following @JohnSundell's talks about this and it's exciting to see the internals now. 🎉 

I went to go play around with this today, so I created a new Xcode project and dropped a Cartfile in, that contained only the following:

```gml
github "spotify/HubFramework" ~> 2.0
```

Upon running `carthage bootstrap`, I got the following result:

```bash
$ carthage bootstrap
*** No Cartfile.resolved found, updating dependencies
*** Cloning HubFramework
*** Checking out HubFramework at "2.0.0"
*** xcodebuild output can be found in /var/folders/23/894nltqn68z80wshzlztjd20393j93/T/carthage-xcodebuild.JgNOeT.log
*** Skipped building HubFramework due to the error:
Dependency "HubFramework" has no shared framework schemes

If you believe this to be an error, please file an issue with the maintainers at https://github.com/spotify/HubFramework/issues/new
```

Looking at the schemes shared from the project, HubFramework is definitely a shared scheme... but it's a static library and not a dynamic framework, as Carthage will expect it to be.

Are there certain motivations around keeping it a static library and not a dynamic framework? If so, you might want to remove the Carthage install instructions, as I believe Carthage has no intention of supporting static libraries at the time being. If not, I can work on a pull request to create a dynamic framework target! Thanks!

rastersize:
Since we don’t expose a dynamic framework target Carthage will error out. This makes sure Carthage can still be used for its dependency resolution.

Also bumps our Gems, most notably xcpretty to 0.2.3 which brings in almost a full year’s worth of fixes.

Should resolve #39 until we have #42.

@spotify/objc-dev @eliperkins


rastersize:
Oh… `scan` doesn’t take a target parameter. That’s unfortunate. I’ll investigate what we can do to make Carthage less errory.

fastlane/Fastfile
```diff
@@ -12,7 +12,7 @@ platform :ios do
     UI.message("Simulator testing device set to '#{device}'")
 
     scan(
-      :workspace => 'HubFramework.xcworkspace',
+      :project => 'HubFramework.xcodeproj',
       :scheme => 'HubFramework',
```
JohnSundell: This can't be updated to `:target`?

rastersize: Unfortunately not. Xcode doesn’t support running tests with just a target name. You need a scheme 😢

rastersize:
I’m going to close this in favor of adding a dynamic framework target instead.

PR #46, opened 2016-09-30
Fix `ignoredAutoEquatablePropertyNames` not compiling for device

rastersize:
The `index` selector reference was ambiguous and the build would fail. This change resolves that by introducing a typed and compile time checked key-path to string macro.

@spotify/objc-dev

### DIFF
--- a/sources/HUBKeyPath.h
+++ b/sources/HUBKeyPath.h
@@ -1,0 +1,35 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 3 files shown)


sources/HUBKeyPath.h
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ *  Returns a key-path string that has been validated at compile-time.
+ *
+ *  ```objc
+ *  HUBKeyPath(self.object, property) => @"property"
+ *  HUBKeyPath(self, object.property) => @"object.property"
+ *  ```
+ *
+ *  In a class method (such as `keyPathsForValuesAffectingvalueForKey:`) you can cast `nil` to a instance pointer
```
JohnSundell: Clever! 👍

PR #47, opened 2016-10-01
View Model: Use HUBKeyPath

JohnSundell:
Use the newly introduced `HUBKeyPath` macro to return ignored auto equatable property names, instead of using `NSStringFromSelector`.


JohnSundell:
@spotify/objc-dev

PR #50, opened 2016-10-03
Add support for UINavigationItem

JohnSundell:
This change adds support for setting a `UINavigationItem` on a `HUBViewModel`, which will then be used when its `HUBViewController` is added to a navigation controller.

This is done by replacing `navigationBarTitle` with `navigationItem` on `HUBViewModel` (since you can now just use the navigation item’s title). No backwards compatibility is added here, since the use cases for using a `HUBViewModel` directly are very small.

However, on the builder (`HUBViewModelBuilder`) backwards compatibility is achieved by letting `navigationBarTitle` be an alias for `navigationItem.title`.

Some equality checking code is extracted from `HUBAutoEquatable` into `HUBUtilities` to avoid code duplication.

### DIFF
--- a/include/HubFramework/HUBViewModel.h
+++ b/include/HubFramework/HUBViewModel.h
@@ -20,9 +20,9 @@
  */
 
 #import "HUBSerializable.h"
-#import <UIKit/UIKit.h>
 
 @protocol HUBComponentModel;
+@class UINavigationItem;
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/include/HubFramework/HUBViewModel.h
+++ b/include/HubFramework/HUBViewModel.h
@@ -20,6 +20,7 @@
  */
 
 #import "HUBSerializable.h"
+#import <UIKit/UIKit.h>
 
 @protocol HUBComponentModel;
 
@@ -47,11 +48,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy, readonly, nullable) NSString *identifier;
 
 /**
- *  The title that should be displayed in the view's navigation bar
- *
- *  The value of this property will be assigned to the view controller's title property
+ *  The navigation item that the view should use when presented in a `UINavigationController`
  */
-@property (nonatomic, copy, readonly, nullable) NSString *navigationBarTitle;
+@property (nonatomic, copy, readonly, nullable) UINavigationItem *navigationItem;
 
 #pragma mark - Component models
 
--- a/include/HubFramework/HUBViewModelBuilder.h
+++ b/include/HubFramework/HUBViewModelBuilder.h
@@ -20,6 +20,7 @@
  */
 
 #import "HUBJSONCompatibleBuilder.h"
+#import <UIKit/UIKit.h>
 
 @protocol HUBComponentModelBuilder;
 
@@ -68,9 +69,17 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  The title that the view should have in the navigation bar
  *
- *  The value of this property will be assigned to the view controller's title property
+ *  This property is an alias for `navigationItem.title`.
  */
 @property (nonatomic, copy, nullable) NSString *navigationBarTitle;
+
+/**
+ *  The navigation item that the view should use when presented in a `UINavigationController`
+ *
+ *  You can use this navigation item to set what title, bar buttons etc that the view's navigation bar
+ *  should contain. Only relevant when the view's controller is added to a container view controller.
+ */
+@property (nonatomic, strong, readonly) UINavigationItem *navigationItem;
 
 /**
  *  The builder to use to build a model for the view's header component
--- a/include/HubFramework/HUBViewModelBuilder.h
+++ b/include/HubFramework/HUBViewModelBuilder.h
@@ -20,9 +20,9 @@
  */
 
 #import "HUBJSONCompatibleBuilder.h"
-#import <UIKit/UIKit.h>
 
 @protocol HUBComponentModelBuilder;
+@class UINavigationItem;
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/sources/HUBUtilities.h
+++ b/sources/HUBUtilities.h
@@ -24,8 +24,29 @@
 #import "HUBComponent.h"
 #import "HUBJSONCompatibleBuilder.h"
 #import "HUBSerializable.h"
+#import "HUBKeyPath.h"
 
 NS_ASSUME_NONNULL_BEGIN
+
+/**
+ *  Return whether two properties are considered equal
+ *
+ *  @param objectA The first object
+ *  @param objectB The second object
+ *  @param propertyName The name of the property to compare
+ *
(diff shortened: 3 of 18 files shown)


JohnSundell:
@spotify/objc-dev

include/HubFramework/HUBViewModelBuilder.h
```diff
@@ -68,11 +69,19 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  The title that the view should have in the navigation bar
  *
- *  The value of this property will be assigned to the view controller's title property
+ *  This property is an alias for `navigationItem.title`.
  */
 @property (nonatomic, copy, nullable) NSString *navigationBarTitle;
```
rastersize: Do we intend for this property to live on? Or do we want to encourage that API users transition to using the `navigationItem` property?

If the latter, how about also deprecating this property like
```objc
DEPRECATED_MSG_ATTRIBUTE("Use the navigationItem property instead")
@property (nonatomic, copy, nullable) NSString *navigationBarTitle;
```

include/HubFramework/HUBViewModel.h
```diff
@@ -20,6 +20,7 @@
  */
 
 #import "HUBSerializable.h"
+#import <UIKit/UIKit.h>
```
rastersize: Do we need to import UIKit here, how about a forward class declaration of `UINavigationItem`?

include/HubFramework/HUBViewModelBuilder.h
```diff
@@ -20,6 +20,7 @@
  */
 
 #import "HUBJSONCompatibleBuilder.h"
+#import <UIKit/UIKit.h>
```
rastersize: Same as for `HUBViewModel.h`, how about a forward class declaration of `UINavigationItem`?

sources/HUBUtilities.h
```diff
@@ -24,10 +24,31 @@
...
 /**
+ *  Return whether two properties are considered equal
+ *
+ *  @param objectA The first object
+ *  @param objectB The second object
+ *  @param propertyName The name of the property to compare
+ *
+ *  Two nil values are considered equal.
+ */
+static inline BOOL HUBPropertyIsEqual(NSObject * _Nullable objectA, NSObject * _Nullable objectB, NSString *propertyName) {
```
rastersize: What’s the reason we need to inline this function? Seems like it could just be a normal function instead, no?

sources/HUBUtilities.h
```diff
@@ -95,6 +116,76 @@ static inline NSDictionary<NSString *, id> * _Nullable HUBMergeDictionaries(NSDi
...
+/**
+ *  Return whether two `UINavigationItem` instances are equal
+ *
+ *  @param navigationItemA The first navigation item
+ *  @param navigationItemB The second navigation item
+ *
+ *  The two instances are considered equal if all properties which have names included in the array
+ *  obtained by calling `HUBNavigationItemPropertyNames()` are equal.
+ */
+static inline BOOL HUBNavigationItemEqualToNavigationItem(UINavigationItem *navigationItemA, UINavigationItem *navigationItemB) {
```
rastersize: What’s the reason we need to inline this function? Seems like it could just be a normal function instead, no?

The overhead of executing the function should be very minimal compared to the complexity of it’s implementation. So might as well have it be a normal function imo.

sources/HUBUtilities.h
```diff
@@ -95,6 +116,76 @@ static inline NSDictionary<NSString *, id> * _Nullable HUBMergeDictionaries(NSDi
...
+ *
+ *  @param navigationItemA The navigation item to copy values into
+ *  @param navigationItemB Any navigation item to copy values from
+ *
+ *  @return navigationItemA
+ *
+ *  If `navigationItemB` is nil, all `navigationItemA` properties will be reset to `nil`. To determine
+ *  which properties that should be handled, `HUBNavigationItemPropertyNames()` is called.
+ */
+static inline UINavigationItem *HUBCopyNavigationItemProperties(UINavigationItem *navigationItemA, UINavigationItem * _Nullable navigationItemB) {
```
rastersize: This is so big and complex it feels like it shouldn’t be inlined imo. How about changing it to a normal function?

include/HubFramework/HUBViewModelBuilder.h
```diff
@@ -68,11 +69,19 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
 @property (nonatomic, copy, nullable) NSString *navigationBarTitle;
 
 /**
+ *  The navigation item that the view should use when presented in a `UINavigationController`
+ *
+ *  You can use this navigation item to set what title, bar buttons etc that the view's navigation bar
+ *  should contain. Only relevant when the view's controller is added to a container view controller.
+ */
+@property (nonatomic, strong, readonly) UINavigationItem *navigationItem;
```
rastersize: Given its implementation, should we mark this as `null_resetable`? Since it would technically be possible to set it to `nil` and have it reset back to a default implementation when accessed again. Only if it’s a use-case we intend to support of course.

include/HubFramework/HUBViewModel.h
JohnSundell: Could do, but does it matter? I mean, `UIKIt` is required by the entire framework.

include/HubFramework/HUBViewModelBuilder.h
JohnSundell: It can't be reset to `nil`, since it's `readonly`. It will only be lazily created, but that's an implementation detail.

sources/HUBUtilities.h
JohnSundell: No reason, just in the habit of doing it. Adding `inline` as no downsides AFAIK though 😄

sources/HUBUtilities.h
JohnSundell: Fair point actually, will change!

include/HubFramework/HUBViewModelBuilder.h
JohnSundell: Yeah, we have convenience APIs like that on `HUBComponentModel` as well, aliases to perform common tasks, such as setting `imageURLs` etc. So no need to deprecate it at this time :)

sources/HUBUtilities.h
Goos: Since it's only used in two places (both of which are in loops), wouldn't it be a perfect candidate for inlining? :smile:

include/HubFramework/HUBViewModelBuilder.h
rastersize: My bad! Missed that keyword 😅 👍

include/HubFramework/HUBViewModel.h
rastersize: It does speed up compile time. Up to you :)

sources/HUBUtilities.h
rastersize: @Goos Right, let’s keep this inline then! 😄

PR #52, opened 2016-10-03
Avoid HUBComponentWrapper state saves before configure

Goos:
Fixes a bug where the HUBComponentWrapper would occasionally save its UI state before having been configured properly.

@spotify/objc-dev


rastersize:
@Goos FYI:
```
[19:42:25]: ▸ ⚠️  /Users/travis/build/spotify/HubFramework/tests/HUBComponentWrapperTests.m:104:5: Null is returned from a method that is expected to return a non-null value
[19:42:25]: ▸ return nil;
[19:42:25]: ▸     ^~~~~~~~~~
```

JohnSundell:
@rastersize Shouldn't our super strict build settings catch that? 😱

Goos:
Yeah I was a bit surprised I was allowed to do that 😮  can I leave it or should we return something there in case future builds pick up the warning?

JohnSundell:
@Goos I'd add an `XCTFail` call in that method and then return `(HUBComponentWrapper *)nil`.

PR #55, opened 2016-10-05
Fix max target content offset after updates

Goos:
Fixes an issue where the collection view layout wouldn't take the bottom content inset into account when determining the proper offset after updates.

@spotify/objc-dev

### DIFF
--- a/sources/HUBCollectionViewLayout.m
+++ b/sources/HUBCollectionViewLayout.m
@@ -208,7 +208,7 @@ NS_ASSUME_NONNULL_BEGIN
     CGFloat const minContentOffset = -self.collectionView.contentInset.top;
     offset.y = MAX(minContentOffset, offset.y);
     // ...or beyond the bottom.
-    CGFloat maxContentOffset = MAX(self.contentSize.height - CGRectGetHeight(self.collectionView.frame), minContentOffset);
+    CGFloat maxContentOffset = MAX(self.contentSize.height + self.collectionView.contentInset.bottom - CGRectGetHeight(self.collectionView.frame), minContentOffset);
     offset.y = MIN(maxContentOffset, offset.y);
     
     self.previousLayoutAttributesByIndexPath = nil;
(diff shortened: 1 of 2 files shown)


JohnSundell:
👍

Goos:
Could you have another look @JohnSundell? Turns out we had a typo in the `targetContentOffsetForProposedContentOffset:`, yay testing!

sources/HUBCollectionViewLayout.m
```diff
@@ -183,7 +183,7 @@ - (CGPoint)targetContentOffsetForProposedContentOffset:(CGPoint)proposedContentO
     
     for (NSIndexPath *indexPath in self.lastViewModelDiff.insertedBodyComponentIndexPaths) {
         if (indexPath.item <= topmostVisibleIndex) {
-            UICollectionViewLayoutAttributes *attributes = self.previousLayoutAttributesByIndexPath[indexPath];
+            UICollectionViewLayoutAttributes *attributes = self.layoutAttributesByIndexPath[indexPath];
```
Goos: The size of the added cells should be computed from the new layout attributes, rather than the previous.

sources/HUBCollectionViewLayout.m
```diff
@@ -198,7 +198,7 @@ - (CGPoint)targetContentOffsetForProposedContentOffset:(CGPoint)proposedContentO
     for (NSIndexPath *indexPath in self.lastViewModelDiff.reloadedBodyComponentIndexPaths) {
         if (indexPath.item < topmostVisibleIndex) {
             UICollectionViewLayoutAttributes *oldAttributes = self.previousLayoutAttributesByIndexPath[indexPath];
-            UICollectionViewLayoutAttributes *newAttributes = self.previousLayoutAttributesByIndexPath[indexPath];
+            UICollectionViewLayoutAttributes *newAttributes = self.layoutAttributesByIndexPath[indexPath];
```
Goos: ...As should the reloads

Issue #19, opened 2016-09-29
Send content offset notifications to child components


JohnSundell:
Currently only root components are able to recieve content offset notifications, by conforming to `HUBComponentContentOffsetObserver`. We should extend this functionality to child components as well.

PR #58, opened 2016-10-06
Content offset updates for child component

Goos:
Adds the possibility of child components listening for content offset updates via the `HUBComponentContentOffsetObserver` protocol.

Also changes slightly how the list of eligible components are maintained for base-level components as well, adding and removing them in the `willDisplayCell:` & `didEndDisplayingCell:` callbacks.

Resolves https://github.com/spotify/HubFramework/issues/19

@spotify/objc-dev

### DIFF
--- a/sources/HUBComponentWrapper.h
+++ b/sources/HUBComponentWrapper.h
@@ -54,7 +54,8 @@ NS_ASSUME_NONNULL_BEGIN
  *  @param childIndex The index of the child component that is about to appear
  */
 - (void)componentWrapper:(HUBComponentWrapper *)componentWrapper
-          childComponent:(HUBComponentWrapper *)childComponent
+          childComponent:(nullable HUBComponentWrapper *)childComponent
+               childView:(UIView *)childComponentView
        willAppearAtIndex:(NSUInteger)childIndex;
 
 /**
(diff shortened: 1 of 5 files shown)


sources/HUBComponentWrapper.m
```diff
@@ -282,8 +281,9 @@ - (void)component:(id<HUBComponentWithChildren>)component didStopDisplayingChild
         return;
     }
     
+    HUBComponentWrapper * const childComponent = self.childrenByIndex[@(childIndex)];
```
JohnSundell: This will break for components that are using normal `UIViews` for their child components instead of creating them through the framework. It's a usecase that we want (and need to - for legacy) support :/

Basically this will be `nil` for those type fo components.

Goos: Oh damn, didn't know that! So would these UIView's also be able to conform to HUBComponentContentObserver, or is it alright if we only add support for content offset observing for the components created through the framework?

JohnSundell: Yeah we need to add some tests to cover that! 😁 Yeah, it's definetly alright to limit the functionality for `UIView`-based components, as long as they continue working as expected 😀

sources/HUBComponentWrapper.h
```diff
@@ -54,7 +54,8 @@ NS_ASSUME_NONNULL_BEGIN
  *  @param childIndex The index of the child component that is about to appear
  */
 - (void)componentWrapper:(HUBComponentWrapper *)componentWrapper
-  childComponentWithView:(UIView *)childComponentView
+          childComponent:(nullable HUBComponentWrapper *)childComponent
```
JohnSundell: Add documentation for this new parameter :)

PR #59, opened 2016-10-06
Set the viewModel after performing batch updates

nataliq:
UICollectionView’s `performBatchUpdates:` method checks the number of items in the collection view before performing the update block and after that. It expects the following: 
```The number of items contained in an existing section after the update must be equal to the number of items contained in that section before the update, plus or minus the number of items inserted or deleted from that section and plus or minus the number of items moved into or out of that section.```
If we update the view model before performing the batch updates `collectionView:numberOfItemsInSection:` will return the updated count even before the actual update and it will fail after that with uncaught exception `'NSInternalInconsistencyException', reason: 'Invalid update’`. 
For example we have 0 items initially and we add 2, before the block we will have 2 items and we will add 2 from the “lastViewModelDiff” so at the end the collection view will expect to have 4 items but it will have only 2.

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -630,6 +630,14 @@ NS_ASSUME_NONNULL_BEGIN
         [self.collectionView reloadData];
         
         [layout computeForCollectionViewSize:self.collectionView.frame.size viewModel:viewModel diff:self.lastViewModelDiff];
+
+        if (self.viewHasAppeared) {
+            /* Forcing a re-layout as the reloadData-call doesn't trigger the numberOfItemsInSection:-calls
+             by itself, and batch update calls don't play well without having an initial item count. */
+            [self.collectionView setNeedsLayout];
+            [self.collectionView layoutIfNeeded];
+        }
+        
         self.lastViewModelDiff = nil;
     } else {
         void (^updateBlock)() = ^{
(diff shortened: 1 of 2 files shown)


Goos:
That sounds odd. AFAIK, the UICollectionView shouldn't ask for a new count before the batch update, it should only compare the latest value returned from `numberOfItemsInSection:`, usually from the previous reload. 

Could you try and see at which points in time it triggers the `numberOfItemsInSection:`-call when this error occurs?

nataliq:
Thanks for the input. I will try to write a test that captures this case, then we can use it even if we update the solution. @JohnSundell I missed that the delegate is called when the viewModel updates, that definitely requires changes. I have 100% reproducible scenario and validated that `numberOfItemsInSection:` is called before and after the batch update but I can try to post more concrete example later. 
Posting some links from people that had similar problem and fixed it with a different approach:
https://fangpenlin.com/posts/2016/04/29/uicollectionview-invalid-number-of-items-crash-issue/
http://stackoverflow.com/questions/37846653/nightmare-with-performbatchupdates-crash

nataliq:
@JohnSundell After discussing with @Goos and applying some of his changes from https://github.com/spotify/HubFramework/pull/61 I think we covered two crashes.

JohnSundell:
@nataliq Or actually, don't worry about it. Just merge once ready :)

nataliq:
@JohnSundell I was thinking about that but I actually want to keep the initial commit description as a reference, we might have similar problems in the future but hopefully not 😀

JohnSundell:
@nataliq Sounds cool! 👍

PR #61, opened 2016-10-07
performBatchUpdates crash fix

Goos:
Attempts to fix a crash that would occur if the initial UICollectionView load was done with a batch update.

@nataliq, @JohnSundell

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -76,6 +76,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable) HUBViewModelDiff *lastViewModelDiff;
 @property (nonatomic, assign) BOOL viewHasAppeared;
 @property (nonatomic, assign) BOOL viewHasBeenLaidOut;
+@property (nonatomic) BOOL hasPerformedInitialReload;
 @property (nonatomic) BOOL viewModelIsInitial;
 @property (nonatomic) BOOL viewModelHasChangedSinceLastLayoutUpdate;
 @property (nonatomic) CGFloat visibleKeyboardHeight;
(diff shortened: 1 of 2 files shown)


sources/HUBViewControllerImplementation.m
```diff
@@ -609,11 +610,18 @@ - (void)reloadCollectionViewWithViewModel:(id<HUBViewModel>)viewModel animated:(
     /* Performing batch updates inbetween viewDidLoad and viewDidAppear is seemingly not allowed, as it
      causes an assertion inside a private UICollectionView method. If no diff exists, fall back to
      a complete reload. */
-    if (!self.viewHasAppeared || self.lastViewModelDiff == nil) {
+    if (!self.hasPerformedInitialReload || !self.viewHasAppeared || self.lastViewModelDiff == nil) {
```
JohnSundell: With this, can we remove `viewHasBeenLaidOut`, or do they complement each other?

JohnSundell:
I'm guessing this is an alternative implementation to https://github.com/spotify/HubFramework/pull/59?

sources/HUBViewControllerImplementation.m
```diff
@@ -609,11 +610,18 @@ - (void)reloadCollectionViewWithViewModel:(id<HUBViewModel>)viewModel animated:(
...
         [self.collectionView reloadData];
         
         [layout computeForCollectionViewSize:self.collectionView.frame.size viewModel:viewModel diff:self.lastViewModelDiff];
+        
+        /* Forcing a re-layout as the reloadData-call doesn't trigger the numberOfItemsInSection:-calls 
+           by itself, and batch update calls don't play well without having an initial item count. */
```
nataliq: Actually I was able to reproduce the crash even after the `numberOfItemsInSection:` was called initially so not sure that this guarantees the expected behaviour.

Goos:
@JohnSundell that's right! We're working on a third one right now :smile:

PR #62, opened 2016-10-07
Add API to create view controllers without registering features

JohnSundell:
This change adds a new way to create view controllers, without having to pre-register a feature with the framework. The use case for this is to more easily be able to create view controller, which decreases the learning curve to get into using the framework.

It also will make it easier to use the Hub Framework for playgrounds.

To make this happen, another convenience API is introduced to be able to create content operation factories using a concrete class that takes a block. Again, this reduces the barrier of entry for creating content operations, and the need for boilerplate.

### DIFF
--- a/include/HubFramework/HUBBlockContentOperationFactory.h
+++ b/include/HubFramework/HUBBlockContentOperationFactory.h
@@ -1,0 +1,26 @@
+#import "HUBContentOperationFactory.h"
+#import "HUBHeaderMacros.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ *  A concrete content opereation factory implementation that uses a block
+ *
+ *  You can use this content operation factory in case you want to implement a simple factory that
+ *  doesn't need any injected dependencies or complex logic. For more advanced use cases, see the
+ *  `HUBContentOperationFactory` protocol, that you can implement in a custom object.
+ */
+@interface HUBBlockContentOperationFactory : NSObject  <HUBContentOperationFactory>
+
+/**
+ *  Initialize an instance of this class with a block that creates content operations
+ *
+ *  @param block The block used to create content operations. The input parameter of the block will
+ *         be the view URI that content operations should be created for. This block will be copied
+ *         and called every time this factory is asked to create content operations.
+ */
+- (instancetype)initWithBlock:(NSArray<id<HUBContentOperation>> *(^)(NSURL *))block HUB_DESIGNATED_INITIALIZER;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/include/HubFramework/HUBBlockContentOperationFactory.h
+++ b/include/HubFramework/HUBBlockContentOperationFactory.h
@@ -1,3 +1,24 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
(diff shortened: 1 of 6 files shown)


JohnSundell:
@spotify/objc-dev

JohnSundell:
Also planning to update the "Getting Started Guide" to use this API instead, since it's so much easier to get into and doesn't require modifying the navigation stack. Will do in a subsequent PR thou :)

include/HubFramework/HUBViewControllerFactory.h
```diff
@@ -46,17 +47,37 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *         Hub Framework's feature registry, it will simply be assigned to the view controller.
+ *  @param contentOperations The content operations to use to load the content for the view controller.
+ *  @param featureIdentifier The identifier of the feature that the view controller will belong to. Will
+ *         be made available to content operations as part of `HUBFeatureInfo`.
+ *  @param featureTitle The title of the feature that the view controller will belong to. Used for its
+ *         default title, and also made available to contnet operations as part of `HUBFeatureInfo`.
+ */
+- (UIViewController<HUBViewController> *)createViewControllerForViewURI:(NSURL *)viewURI
+                                                      contentOperations:(NSArray<id<HUBContentOperation>> *)contentOperations
+                                                      featureIdentifier:(NSString *)featureIdentifier
```
Goos: If the point of this method is to enable creating view controllers without registering a feature, does it still make sense to require a feature identifier?

include/HubFramework/HUBBlockContentOperationFactory.h
```diff
@@ -0,0 +1,47 @@
...
+@interface HUBBlockContentOperationFactory : NSObject  <HUBContentOperationFactory>
+
+/**
+ *  Initialize an instance of this class with a block that creates content operations
+ *
+ *  @param block The block used to create content operations. The input parameter of the block will
+ *         be the view URI that content operations should be created for. This block will be copied
+ *         and called every time this factory is asked to create content operations.
+ */
+- (instancetype)initWithBlock:(NSArray<id<HUBContentOperation>> *(^)(NSURL *))block HUB_DESIGNATED_INITIALIZER;
```
Goos: Perhaps we could typedef this block to make it more readable?

include/HubFramework/HUBViewControllerFactory.h
JohnSundell: Yeah, it's needed for other APIs, such as passing `HUBFeatureInfo` to content operations. Better to require it to enable content operations to be reusable between view controllers created this way and those created from feature registrations.

Goos:
Awesome stuff! 🎉

include/HubFramework/HUBBlockContentOperationFactory.h
JohnSundell: Good suggestion, I'll do that! 👍

tests/HUBViewControllerFactoryTests.m
```diff
@@ -103,6 +103,30 @@ - (void)testCreatingViewControllerForInvalidViewURIReturnsNil
...
+    UIViewController<HUBViewController> * const viewController = [self.manager.viewControllerFactory createViewControllerForViewURI:viewURI
+                                                                                                                  contentOperations:@[contentOperation]
+                                                                                                                  featureIdentifier:@"identifier"
+                                                                                                                       featureTitle:@"Title"];
+    
+    [viewController viewWillAppear:NO];
+    
+    XCTAssertTrue(contentOperationCalled);
+    XCTAssertEqualObjects(viewController.featureIdentifier, @"identifier");
+    XCTAssertEqualObjects(viewController.navigationItem.title, @"Title");
```
nataliq: What about checking the feature registry for a registration for the provided URI?

nataliq: Never mind, I was under the impression that there is a feature registration behind the scenes and my suggestion was checking that it is not nil. My bad :)

PR #64, opened 2016-10-10
Fix content offset adjustment error when view is empty

JohnSundell:
This patch fixes a bug that would cause the content offset of a view controller to be incorrect after an initial update, where the view goes from being empty to being populated with components.

The problem was that the topmost visible index would end up being `NSNotFound`, which would make the view jump to the bottom.


JohnSundell:
@spotify/objc-dev

FranGamb:
🙌🎉🎉

PR #65, opened 2016-10-10
Notify header component of content offset change when it appears

JohnSundell:
This patch fixes a bug where header components wouldn’t be notified of a content offset change when they appear. Body & overlay components already had this intended behavior.

The fix is to merge the “component will appear” logic into a single method, to make sure that all components are treated the same in this regard.


JohnSundell:
@spotify/objc-dev

PR #66, opened 2016-10-10
Umbrella header: Add HUBBlockContentOperationFactory

JohnSundell:
Missed to add it to the umbrella header when it was introduced.


JohnSundell:
@spotify/objc-dev

PR #68, opened 2016-10-11
Support nested touchable views inside a UI[Table/Collection]ViewCell

JohnSundell:
Because of the ways UIKit handles touches in `UITableViewCell` and `UICollectionViewCell`, and we want to support using “orphaned” cells as component views - we need to employ some custom touch handling.

It turns out though, that the custom touch handling wasn’t taking nested touchable views (such as a `UIButton` inside a cell) into account. This has now been fixed, by making sure we only do custom handling when the content view of a cell was touched.

### DIFF
--- a/sources/HUBComponentCellWrapperView.m
+++ b/sources/HUBComponentCellWrapperView.m
@@ -36,6 +36,22 @@
 
 - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event
 {
+    UIView * const viewAtPoint = [self.componentView hitTest:point withEvent:event];
+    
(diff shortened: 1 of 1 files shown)


JohnSundell:
@spotify/objc-dev

JohnSundell:
@Goos It's awesome that you're "that guy" 😄 We should always focus on a really high test coverage - but in this case it becomes pretty much impossible to unit test - since it relies on UIKit's internal hit testing mechanisms, that won't get triggered in a unit test 😢

Goos:
Aha gotcha :smile: no worries

PR #71, opened 2016-10-12
Fix bug with incorrect order of components with preferred index.

mvlasieva:
The bug was only reproducible if there are 4 or more components with preferred index and some components without it, and manifested in incorrect resulting order.  
Fixed by sorting components by preferred index before inserting them at required positions.

@Goos

### DIFF
--- a/sources/HUBComponentModelBuilderImplementation.m
+++ b/sources/HUBComponentModelBuilderImplementation.m
@@ -94,8 +94,9 @@ NS_ASSUME_NONNULL_BEGIN
         
         [sortedBuilders addObject:builder];
     }
-    
-    for (NSNumber * const preferredIndex in buildersByPreferredIndex) {
+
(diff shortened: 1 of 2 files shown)


sources/HUBComponentModelBuilderImplementation.m
```diff
@@ -94,8 +94,9 @@ @implementation HUBComponentModelBuilderImplementation
         
         [sortedBuilders addObject:builder];
     }
-    
-    for (NSNumber * const preferredIndex in buildersByPreferredIndex) {
+
+    NSArray *sortedPreferredIndexes = [[buildersByPreferredIndex allKeys] sortedArrayUsingSelector: @selector(compare:)];
```
JohnSundell: 🎨 Don't include a space between `:` and `@selector`.

JohnSundell: @mvlasieva awesome, thanks! 👍

JohnSundell:
@mvlasieva Merge `master` into your branch, then I can merge :)

PR #72, opened 2016-10-12
Remove HUBComponentImageDataBuilderImplementation from test target

JohnSundell:
Should only be included in library target.


JohnSundell:
@spotify/objc-dev

Issue #51, opened 2016-10-03
Image loader: Replace cache-check with timer


JohnSundell:
Currently, we ask each implementation of `HUBImageLoader` to return whether an image was loaded from cache or not. This parameter is in turn used to decide whether a component should animate a change of image.

Instead of doing this, let the user of `HUBImageLoader` (currently `HUBViewControllerImplementation`) keep track of how long time the image download took, and then - if the value exceeded a certain threshold (in the Spotify app we use `0.07` seconds, which seems to be a reasonable value). This removes the need for image loaders to perform this work, and guarantees a consistent behavior irregardless of image loader implementation.

When this is done, we can remove the `loadedFromCache` parameter in `HUBImageLoaderDelegate`.

JohnSundell:
@Goos want to take care of this? 😄

krris:
Hey! I will work on this.

krris:
Implementation is done in https://github.com/spotify/HubFramework/pull/75

PR #75, opened 2016-10-13
Image loader: Replace cache-check with timer

krris:
This PR resolves an issue https://github.com/spotify/HubFramework/issues/51


sources/HUBViewControllerImplementation.m
```diff
@@ -48,6 +48,8 @@
 #import "HUBActionHandler.h"
 #import "HUBViewModelDiff.h"
 
+CFTimeInterval const kDownloadingTimeTreshold = 0.07;
```
krris: Shouldn't we expose this value to the user? What do you think?

sources/HUBComponentImageLoadingContext.h
```diff
@@ -40,6 +40,8 @@ NS_ASSUME_NONNULL_BEGIN
 /// Any index of a child component that the image is for
 @property (nonatomic, copy, readonly, nullable) NSNumber *childIndex;
 
+@property (nonatomic, assign, readonly) CFTimeInterval timestamp;
```
rastersize: Please document this property.

sources/HUBComponentImageLoadingContext.h
```diff
@@ -51,7 +53,8 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithImageType:(HUBComponentImageType)imageType
                   imageIdentifier:(nullable NSString *)imageIdentifier
                 wrapperIdentifier:(NSUUID *)wrapperIdentifier
-                       childIndex:(nullable NSNumber *)childIndex HUB_DESIGNATED_INITIALIZER;
+                       childIndex:(nullable NSNumber *)childIndex
+                        timestamp:(CFTimeInterval)timestamp HUB_DESIGNATED_INITIALIZER;
```
rastersize: Please add parameter documentation for `timestamp`.

rastersize: How about changing the type of this parameter and the associated property to `NSTimeInterval`? That would keep the API cleaner, abstracting away that we currently use Core Animation function. Should make things nicer to use from Swift.

sources/HUBViewControllerImplementation.m
rastersize: Please rename it to `HUBDownloadTimeThreshold` to adhere to our style guidelines:

Only pure C constants used by pure C code should be prefixed with a `k`. All other constant, and symbols, should follow normal Cocoa naming guidelines. I.e. use the symbol prefix (`HUB`).

rastersize: I agree that it probably makes sense to expose this value via some header.

If not we need to add `static` here to not pollute the global namespace, like:
```objc
static CFTimeInterval  const HUBDownloadingTimeThreshold = 0.07;
```

sources/HUBComponentImageLoadingContext.h
JohnSundell: Yeah, totally agree - I think `NSTimeInterval` should be used.

sources/HUBViewControllerImplementation.m
JohnSundell: Why does it need to be exposed? Is there any API use case for it? Otherwise I think it's better to keep it internal.

JohnSundell: Also make it `static`. And since it's only relevant for images add that information to the variable name. Like this:
`HUBImageDownloadTimeThreshold`.

sources/HUBViewControllerImplementation.m
```diff
@@ -956,14 +958,15 @@ - (void)loadImageFromData:(id<HUBComponentImageData>)imageData
...
         return;
     }
-    
+
     HUBComponentImageLoadingContext * const context = [[HUBComponentImageLoadingContext alloc] initWithImageType:imageData.type
                                                                                                  imageIdentifier:imageData.identifier
                                                                                                wrapperIdentifier:componentWrapper.identifier
-                                                                                                      childIndex:childIndex];
+                                                                                                      childIndex:childIndex
+                                                                                                       timestamp:CACurrentMediaTime()];
```
JohnSundell: Isn't it better to use `[NSDate date].timeIntervalSinceReferenceDate`, to avoid the `CoreAnimation` dependency?

JohnSundell: It doesn't sync with the server, as in the network right? I think it's fine to rely on `NSDate`.

krris:
@JohnSundell 🎾

tests/HUBViewControllerTests.m
```diff
@@ -439,18 +439,64 @@ - (void)testReloadingImage
...
+    [self simulateViewControllerLayoutCycle];
+
+    id<UICollectionViewDataSource> const collectionViewDataSource = self.collectionView.dataSource;
+    id<HUBImageLoaderDelegate> const imageLoaderDelegate = self.imageLoader.delegate;
+    
+    NSIndexPath * const indexPath = [NSIndexPath indexPathForItem:0 inSection:0];
+    self.collectionView.cells[indexPath] = [collectionViewDataSource collectionView:self.collectionView cellForItemAtIndexPath:indexPath];
+
+    NSTimeInterval downloadFromNetworkTime = 2;
+    [NSThread sleepForTimeInterval:downloadFromNetworkTime];
```
JohnSundell: Clever!

tests/mocks/HUBComponentMock.h
```diff
@@ -91,6 +91,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// Whether the component should act like it is a content offset observer or not
 @property (nonatomic) BOOL isContentOffsetObserver;
 
+/// Whether the component's image was recently animated
+@property (nonatomic) BOOL isImageAnimated;
```
JohnSundell: Nitpick: Make this `readonly`, so that it can't be accidentially set from the outside, and the override it as `readwrite` in the implementation.

JohnSundell: Also since it refers to the past, how about `imageWasAnimated`? (just a suggestion)

krris: Right, I'll make these changes. 👍

krris:
@JohnSundell 🎾

PR #76, opened 2016-10-13
Fixes an issue with multiple updates happening while a view isn't visible

Goos:
The view controller would only keep a reference to the last diff, possibly losing certain earlier updates. This change keeps track of each change and performs them all in the next update.

@spotify/objc-dev

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -73,7 +73,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, readonly) HUBComponentUIStateManager *componentUIStateManager;
 @property (nonatomic, strong, readonly) HUBComponentReusePool *childComponentReusePool;
 @property (nonatomic, strong, nullable) id<HUBViewModel> viewModel;
-@property (nonatomic, strong, nullable) HUBViewModelDiff *lastViewModelDiff;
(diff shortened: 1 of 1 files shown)


sources/HUBViewControllerImplementation.m
```diff
@@ -73,7 +73,7 @@ @interface HUBViewControllerImplementation () <HUBViewModelLoaderDelegate, HUBIm
 @property (nonatomic, strong, readonly) HUBComponentUIStateManager *componentUIStateManager;
 @property (nonatomic, strong, readonly) HUBComponentReusePool *childComponentReusePool;
 @property (nonatomic, strong, nullable) id<HUBViewModel> viewModel;
-@property (nonatomic, strong, nullable) HUBViewModelDiff *lastViewModelDiff;
+@property (nonatomic, strong, nullable) NSArray<HUBViewModelDiff *> *pendingDiffs;
```
JohnSundell: Seems like this never becomes `nil`, so why make it `nullable`?

JohnSundell: Also, looking at the rest of the code, doesn't it make more sense for it to be a `NSMutableArray`?

JohnSundell:
@Goos Can this be closed, now that you have https://github.com/spotify/HubFramework/pull/89?

PR #84, opened 2016-10-19
Add basic indexPath support to the component model.

cerihughes:
This PR:

- Allows a component model to return an `NSIndexPath` representing where it sits in the component hierarchy,
- Updates the `HUBViewController` to return the index paths of all visible body components, not just the top level body components.

### DIFF
--- a/include/HubFramework/HUBComponentModel.h
+++ b/include/HubFramework/HUBComponentModel.h
@@ -229,6 +229,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable, readonly) NSArray<id<HUBComponentModel>> *children;
 
 /**
+ *  The index path of this component in the `HUBComponentModel` hierarchy.
+ */
+@property (nonatomic, strong, readonly) NSIndexPath *indexPath;
+
+/**
  *  Return a child component model for a given index, or `nil` if an invalid index was supplied
  *
  *  @param childIndex The index to return a child component model for
--- a/include/HubFramework/HUBComponentModel.h
+++ b/include/HubFramework/HUBComponentModel.h
@@ -230,6 +230,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  *  The index path of this component in the `HUBComponentModel` hierarchy.
+ *
+ *  @discussion This is a computed property so shouldn't be accessed multiple times in the same block.
  */
 @property (nonatomic, strong, readonly) NSIndexPath *indexPath;
 
--- a/sources/HUBComponentModelImplementation.m
+++ b/sources/HUBComponentModelImplementation.m
@@ -190,6 +190,31 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - HUBComponentModel
 
+- (NSIndexPath *)indexPath
+{
+    // Since we're traversing from this model through the parents, we will record the indices in reverse order.
+    // Capture them in a mutable array for easy reversal.
+    NSMutableArray<NSNumber *> *reversedIndexPath = [NSMutableArray array];
+    [reversedIndexPath addObject:@(self.index)];
+
+    id<HUBComponentModel> parent = self.parent;
+    while (parent) {
+        [reversedIndexPath addObject:@(parent.index)];
+        parent = parent.parent;
+    }
+
+    NSIndexPath *indexPath = nil;
+    for  (NSNumber *index in reversedIndexPath.reverseObjectEnumerator) {
+        if (indexPath == nil) {
+            indexPath = [NSIndexPath indexPathWithIndex:index.unsignedIntegerValue];
+        } else {
+            indexPath = [indexPath indexPathByAddingIndex:index.unsignedIntegerValue];
+        }
+    }
+
+    return indexPath;
+}
+
 - (nullable id<HUBComponentModel>)childAtIndex:(NSUInteger)childIndex
 {
     if (childIndex >= self.children.count) {
--- a/sources/HUBComponentModelImplementation.m
+++ b/sources/HUBComponentModelImplementation.m
@@ -194,17 +194,17 @@ NS_ASSUME_NONNULL_BEGIN
 {
     // Since we're traversing from this model through the parents, we will record the indices in reverse order.
     // Capture them in a mutable array for easy reversal.
-    NSMutableArray<NSNumber *> *reversedIndexPath = [NSMutableArray array];
-    [reversedIndexPath addObject:@(self.index)];
+    NSMutableArray<NSNumber *> *reversedIndices = [NSMutableArray array];
+    [reversedIndices addObject:@(self.index)];
 
     id<HUBComponentModel> parent = self.parent;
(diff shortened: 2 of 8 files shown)


include/HubFramework/HUBComponentModel.h
```diff
@@ -229,6 +229,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable, readonly) NSArray<id<HUBComponentModel>> *children;
 
 /**
+ *  The index path of this component in the `HUBComponentModel` hierarchy.
+ */
+@property (nonatomic, strong, readonly) NSIndexPath *indexPath;
```
JohnSundell: We should add a note that this property is computed, so that it's not accessed multiple times within the same block of code. Alternatively, we could cache the value. It's up to you :)

sources/HUBComponentModelImplementation.m
```diff
@@ -190,6 +190,31 @@ - (NSString *)debugDescription
...
 
+- (NSIndexPath *)indexPath
+{
+    // Since we're traversing from this model through the parents, we will record the indices in reverse order.
+    // Capture them in a mutable array for easy reversal.
+    NSMutableArray<NSNumber *> *reversedIndexPath = [NSMutableArray array];
+    [reversedIndexPath addObject:@(self.index)];
+
+    id<HUBComponentModel> parent = self.parent;
+    while (parent) {
```
JohnSundell: 🎨 Add an explicit comparison to `nil` (optional)

sources/HUBComponentModelImplementation.m
```diff
@@ -190,6 +190,31 @@ - (NSString *)debugDescription
 
 #pragma mark - HUBComponentModel
 
+- (NSIndexPath *)indexPath
+{
+    // Since we're traversing from this model through the parents, we will record the indices in reverse order.
+    // Capture them in a mutable array for easy reversal.
+    NSMutableArray<NSNumber *> *reversedIndexPath = [NSMutableArray array];
```
JohnSundell: Since this is not an index path, shouldn't we name it something like `reversedIndexes`? (just a thought)

sources/HUBComponentWrapper.h
```diff
@@ -158,6 +158,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)saveComponentUIState;
 
+/**
+ *  Returns an array of all child component wrappers that are currently being displayed.
+ *
+ *  @discussion Only direct descendents will be returned. Visible children of children won't be returned.
+ */
+- (NSArray<HUBComponentWrapper *> *)visibleChildren;
```
JohnSundell: Should this me made into a `readonly @property` instead? Since it's not an action, but rather data you can access.

sources/HUBComponentWrapper.m
```diff
@@ -109,6 +109,18 @@ - (void)saveComponentUIState
     }
 }
 
+- (NSArray<HUBComponentWrapper *> *)visibleChildren
+{
+    NSMutableArray<HUBComponentWrapper *> *visibleChildren = [NSMutableArray array];
+    for (NSNumber *visibleViewID in self.visibleChildViewsByIndex) {
```
JohnSundell: It's not an ID, it's an index :)

sources/HUBComponentWrapper.m
```diff
@@ -109,6 +109,18 @@ - (void)saveComponentUIState
...
+        HUBComponentWrapper *childComponentWrapper = self.childrenByIndex[visibleViewID];
+        if (childComponentWrapper) {
```
JohnSundell: 🎨 Compare explicitly to `nil`? (optional)

sources/HUBViewControllerImplementation.m
```diff
@@ -241,16 +241,30 @@ - (void)didReceiveMemoryWarning
...
 {
-    NSMutableDictionary<NSNumber *, UIView *> * const visibleViews = [NSMutableDictionary new];
-    
+    NSMutableDictionary<NSIndexPath *, UIView *> * const visibleViewIndexPaths = [NSMutableDictionary new];
+    NSMutableArray<HUBComponentWrapper *> *visibleComponents = [NSMutableArray array];
+
     for (HUBComponentCollectionViewCell * const cell in self.collectionView.visibleCells) {
         HUBComponentWrapper * const wrapper = [self componentWrapperFromCell:cell];
-        visibleViews[@(wrapper.model.index)] = HUBComponentLoadViewIfNeeded(wrapper);
+        [self addComponentWrapper:wrapper toArray:visibleComponents];
```
JohnSundell: Since this is acessed only from a single place, why does it need to be a method? (it could be made into a function in `HUBUtilities` if you think it warrants it).

tests/HUBComponentModelTests.m
```diff
@@ -159,10 +159,37 @@ - (void)testChildWithIdentifier
     XCTAssertNil([parent childWithIdentifier:@"noChild"]);
 }
 
+- (void)testIndexPaths
+{
+    HUBComponentModelImplementation * const parent = [self createComponentModelWithIdentifier:@"parent" index:0];
+    HUBComponentModelImplementation * const childA = [self createComponentModelWithIdentifier:@"childA" index:0 parent:parent];
+    HUBComponentModelImplementation * const childB = [self createComponentModelWithIdentifier:@"childB" index:1 parent:parent];
+    HUBComponentModelImplementation * const grandchild = [self createComponentModelWithIdentifier:@"grandchild" index:0 parent:childB];
```
JohnSundell: Grandchild! Nice 😆

include/HubFramework/HUBComponentModel.h
cerihughes: I'll add a comment...

sources/HUBComponentModelImplementation.m
cerihughes: Will do.

sources/HUBComponentWrapper.h
cerihughes: Will do.

sources/HUBComponentWrapper.m
cerihughes: Ooops. It *used* to be an ID before I refactored...

sources/HUBViewControllerImplementation.m
```diff
@@ -241,16 +241,30 @@ - (void)didReceiveMemoryWarning
...
+        visibleViewIndexPaths[indexPath] = HUBComponentLoadViewIfNeeded(visibleComponent);
+    }
+
+    return [visibleViewIndexPaths copy];
+}
+
+- (void)addComponentWrapper:(HUBComponentWrapper *)componentWrapper toArray:(NSMutableArray<HUBComponentWrapper *> *)array
+{
+    [array addObject:componentWrapper];
+    for (HUBComponentWrapper *childComponentWrapper in [componentWrapper visibleChildren]) {
```
JohnSundell: 🎨 Now that this is a property, it should be accessed with dot notation.

PR #85, opened 2016-10-19
Separate UICollectionView update logic

Goos:
Separates the UICollectionView update logic into a separate class – the HUBViewModelRenderer. Also moves the diffing of view models to right before the reload in order to avoid unpredictable behaviour caused by multiple updates happening inbetween a UICollectionView reload.

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -46,7 +46,7 @@
 #import "HUBActionContextImplementation.h"
 #import "HUBActionRegistry.h"
 #import "HUBActionHandler.h"
-#import "HUBViewModelDiff.h"
+#import "HUBViewModelRenderer.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -62,6 +62,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable, readonly) id<HUBContentReloadPolicy> contentReloadPolicy;
 @property (nonatomic, strong, nullable, readonly) id<HUBImageLoader> imageLoader;
 @property (nonatomic, strong, nullable) UICollectionView *collectionView;
+@property (nonatomic, strong, nullable) HUBViewModelRenderer *viewModelRenderer;
 @property (nonatomic, assign) BOOL collectionViewIsScrolling;
 @property (nonatomic, strong, readonly) NSMutableSet<NSString *> *registeredCollectionViewCellReuseIdentifiers;
 @property (nonatomic, strong, readonly) NSMutableDictionary<NSURL *, NSMutableArray<HUBComponentImageLoadingContext *> *> *componentImageLoadingContexts;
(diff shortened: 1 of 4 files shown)


Goos:
Adding `do not merge` as it isn't tested yet

sources/HUBViewControllerImplementation.m
```diff
@@ -606,67 +601,30 @@ - (void)createCollectionViewIfNeeded
...
-            }];
-        };
-        
-        if (animated) {
-            updateBlock();
-        } else {
-            [UIView performWithoutAnimation:updateBlock];
-        }
+    if (self.viewModelRenderer == nil) {
+        UICollectionView * const nonnullCollectionView = self.collectionView;
```
JohnSundell: Why not create them both at the same time?

sources/HUBViewModelRenderer.h
```diff
@@ -0,0 +1,55 @@
...
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import <UIKit/UIKit.h>
+#import "HUBViewModel.h"
```
JohnSundell: Use `@protocol`

sources/HUBViewModelRenderer.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param collectionView The collection view to use for rendering.
+ */
+- (instancetype)initWithCollectionView:(UICollectionView *)collectionView HUB_DESIGNATED_INITIALIZER;
+
+/** 
+ * Renders the provided view model in the collection view.
+ * 
+ * @param viewModel The view model to render.
+ * @param usingBatchUpdates @c YES if the renderer should render using batch updates, @c NO otherwise.
```
JohnSundell: This could be simplified as `Whether the renderer should render using batch updates`.

sources/HUBViewModelRenderer.m
```diff
@@ -0,0 +1,93 @@
...
+    }
+    return self;
+}
+
+- (void)renderViewModel:(id<HUBViewModel>)viewModel
+      usingBatchUpdates:(BOOL)usingBatchUpdates
+               animated:(BOOL)animated
+             completion:(void (^)())completionBlock
+{
+    HUBViewModelDiff *diff;
```
JohnSundell: Give this an initial value of `nil`, so we don't end up using an uninitialized variable.

tammofreese: This is not needed in ARC, as object pointers are initialized to nil.

JohnSundell: Ah cool, didn't know that! Thanks @tammofreese! 👍

tammofreese: No worries! :)

PR #87, opened 2016-10-20
View controller: Un-rename visible component views API & improve docs

JohnSundell:
This change changes the name of the API used to retrieve the currently visible body component views back to the name that was used before https://github.com/spotify/HubFramework/pull/84. We don’t need to change the name of the API because the key type was changed.

This change also adds a bit more thorough documentation for the API, explaining what the API user can expect from the index paths used as keys.

### DIFF
--- a/include/HubFramework/HUBViewController.h
+++ b/include/HubFramework/HUBViewController.h
@@ -137,8 +137,8 @@ NS_ASSUME_NONNULL_BEGIN
  *  The views of the body components that are currently visible on screen, keyed by their index path
  *
  *  The index paths used for keys contains the indexes for the components' views, starting from the root. For example,
- *  a root body component at index 4 will have an index path with just the index 4, while the 3rd child of that component
- *  will have an index path with the indexes 4 and 2.
+ *  a root body component at index 4 will have an index path with just the index 4, while the child at index 2 of that
+ *  component will have an index path with the indexes 4 and 2.
  */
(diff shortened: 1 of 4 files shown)


JohnSundell:
@spotify/objc-dev @cerihughes

cerihughes:
🎉

include/HubFramework/HUBViewController.h
```diff
@@ -133,8 +133,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, nullable, readonly) id<HUBViewModel> viewModel;
 
-/// The views of the root body components that are currently visible on screen, keyed by their index paths
-@property (nonatomic, strong, readonly) NSDictionary<NSIndexPath *, UIView *> *visibleBodyComponentViewIndexPaths;
+/**
+ *  The views of the body components that are currently visible on screen, keyed by their index path
+ *
+ *  The index paths used for keys contains the indexes for the components' views, starting from the root. For example,
+ *  a root body component at index 4 will have an index path with just the index 4, while the 3rd child of that component
```
Goos: Wording is kind of confusing as the first example says "index 4", whereas the second example "3rd child". Perhaps replace 3rd child with "child at index 2"? :smile:

KeenanCassidy:
👍

include/HubFramework/HUBViewController.h
JohnSundell: Great point - will fix!

PR #88, opened 2016-10-20
Simplify the generation of the index path.

cerihughes:
Instead of storing the indices in reverse order, then iterating in reverse, store them in the order they'll be accessed in.

Also gets around a static analyser warning, which claims we could return `nil` from the method (which wasn't strictly true).

### DIFF
--- a/sources/HUBComponentModelImplementation.m
+++ b/sources/HUBComponentModelImplementation.m
@@ -192,24 +192,19 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSIndexPath *)indexPath
 {
-    // Since we're traversing from this model through the parents, we will record the indices in reverse order.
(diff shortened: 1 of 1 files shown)


cerihughes:
@krris : This gets rid of the static analyser warning.

Goos:
Could we also avoid boxing them by using an `NSIndexSet`? We would still have to use reverse enumeration for them then though as they are implicitly ordered

JohnSundell:
@cerihughes The conflict is because of me (sorry!) 😛 I added a workaround for the static analyzer warning here: https://github.com/spotify/HubFramework/pull/87/commits/0550094089dc8e7dca5a936221bdae28008a45e8. But I like this approach better! So feel free to pick your implementation when rebasing 😄

JohnSundell:
Wooops, accidental close 😢

PR #89, opened 2016-10-20
Separate UICollectionView update logic and diff only before reloads

Goos:
Separates the UICollectionView update logic into a separate class – the HUBViewModelRenderer. Also moves the diffing of view models to right before the reload in order to avoid unpredictable behaviour caused by multiple updates happening inbetween a UICollectionView reload.

@spotify/objc-dev

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -49,6 +49,7 @@
 #import "HUBActionPerformer.h"
 #import "HUBViewModelDiff.h"
 #import "HUBComponentGestureRecognizer.h"
+#import "HUBViewModelRenderer.h"
 
 static NSTimeInterval const HUBImageDownloadTimeThreshold = 0.07;
 
@@ -74,6 +75,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable, readonly) id<HUBContentReloadPolicy> contentReloadPolicy;
 @property (nonatomic, strong, nullable, readonly) id<HUBImageLoader> imageLoader;
 @property (nonatomic, strong, nullable) UICollectionView *collectionView;
+@property (nonatomic, strong, nullable) HUBViewModelRenderer *viewModelRenderer;
 @property (nonatomic, assign) BOOL collectionViewIsScrolling;
 @property (nonatomic, strong, readonly) NSMutableSet<NSString *> *registeredCollectionViewCellReuseIdentifiers;
 @property (nonatomic, strong, readonly) NSMutableDictionary<NSURL *, NSMutableArray<HUBComponentImageLoadingContext *> *> *componentImageLoadingContexts;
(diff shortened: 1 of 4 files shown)


sources/HUBViewControllerImplementation.m
```diff
@@ -761,67 +757,30 @@ - (void)createCollectionViewIfNeeded
...
+        self.viewModelRenderer = [[HUBViewModelRenderer alloc] initWithCollectionView:nonnullCollectionView];
     }
+
+    [self saveStatesForVisibleComponents];
+
+    [self.viewModelRenderer renderViewModel:viewModel
+                          usingBatchUpdates:self.viewHasAppeared
+                                   animated:animated
+                                 completion:^{
+        [self.delegate viewControllerDidFinishRendering:self];
```
JohnSundell: This is so nice and clean :)

sources/HUBViewModelRenderer.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @param collectionView The collection view to use for rendering.
+ */
+- (instancetype)initWithCollectionView:(UICollectionView *)collectionView HUB_DESIGNATED_INITIALIZER;
+
+/** 
+ * Renders the provided view model in the collection view.
+ * 
+ * @param viewModel The view model to render.
+ * @param usingBatchUpdates @c YES if the renderer should render using batch updates, @c NO otherwise.
```
JohnSundell: You can simplify this by just saying `Whether the renderer should use batch updates`

sources/HUBViewModelRenderer.h
```diff
@@ -0,0 +1,55 @@
...
+ * @param animated @c YES if the renderer should render with animations, @c NO otherwise.
```
JohnSundell: `Whether the renderer should use animations`?

sources/HUBViewModelRenderer.m
```diff
@@ -0,0 +1,101 @@
...
+#import "HUBViewModelRenderer.h"
+#import "HUBViewModelDiff.h"
+#import "HUBCollectionViewLayout.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface HUBViewModelRenderer ()
+
+@property (nonatomic, strong, readonly) UICollectionView *collectionView;
+@property (nonatomic, strong) id<HUBViewModel> lastRenderedViewModel;
```
JohnSundell: `nullable`?

sources/HUBViewModelRenderer.h
Goos: Ugh, yeah, writing javascript gives you the annoying habit of over-clarifying boolean parameters 😅

JohnSundell:
@Goos Can the `do not merge` label be removed, or does this require some manual verification?

Goos:
@JohnSundell I think it should be fine as long as the CI goes through!

Issue #82, opened 2016-10-19
Send component's view to view controller's delegate instead of collection view cell


JohnSundell:
Currently, we send the wrapping `HUBComponentCollectionViewCell` as the `view` parameter when calling methods on `HUBViewControllerDelegate`. We should change this to send the actual component view instead. 

**Current behavior:**

```objective-c
- (void)viewController:(UIViewController<HUBViewController> *)viewController
    componentWithModel:(id<HUBComponentModel>)componentModel
      willAppearInView:(UIView *)componentView
{
    // componentView is always a HUBComponentCollectionViewCell
}
```

**New behavior:**

(Imagine a component with a `UITableViewCell` as its `view`)

```objective-c
- (void)viewController:(UIViewController<HUBViewController> *)viewController
    componentWithModel:(id<HUBComponentModel>)componentModel
      willAppearInView:(UIView *)componentView
{
    // componentView is a UITableViewCell
}
```

JohnSundell:
@krris Wanna take this on? Let me know if you need more context 🙂

krris:
I assign it to myself. I'll get back to you with questions when I start to work on this.

JohnSundell:
@krris Perfect 👍

JohnSundell:
Implemented in https://github.com/spotify/HubFramework/pull/90

PR #90, opened 2016-10-20
Send component's view instead of a collection view cell in HUBViewControllerDelegate.

krris:
Solves https://github.com/spotify/HubFramework/issues/82

Send unwrapped `HUBComponentCollectionViewCell` as the view parameter when calling methods on `HUBViewControllerDelegate`.

### DIFF
--- a/sources/HUBUtilities.h
+++ b/sources/HUBUtilities.h
@@ -55,7 +55,7 @@ static inline BOOL HUBPropertyIsEqual(NSObject * _Nullable objectA, NSObject * _
  *
  *  This function asserts that a view has been loaded after -loadView was sent to the component.
  */
-static inline UIView *HUBComponentLoadViewIfNeeded(id<HUBComponent> _Nullable component) {
+static inline UIView *HUBComponentLoadViewIfNeeded(id<HUBComponent> component) {
     if (component.view == nil) {
         [component loadView];
     }
--- a/sources/HUBUtilities.h
+++ b/sources/HUBUtilities.h
@@ -55,7 +55,7 @@ static inline BOOL HUBPropertyIsEqual(NSObject * _Nullable objectA, NSObject * _
  *
  *  This function asserts that a view has been loaded after -loadView was sent to the component.
  */
-static inline UIView *HUBComponentLoadViewIfNeeded(id<HUBComponent> component) {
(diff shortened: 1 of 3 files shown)


sources/HUBUtilities.h
```diff
@@ -55,7 +55,7 @@ static inline BOOL HUBPropertyIsEqual(NSObject * _Nullable objectA, NSObject * _
  *
  *  This function asserts that a view has been loaded after -loadView was sent to the component.
  */
-static inline UIView *HUBComponentLoadViewIfNeeded(id<HUBComponent> component) {
+static inline UIView *HUBComponentLoadViewIfNeeded(id<HUBComponent> _Nullable component) {
```
JohnSundell: Do we need to make this change? (This function is being used in many places, and it doesn't make a lot of sense to call it with `nil`)

sources/HUBViewControllerImplementation.m
```diff
@@ -963,7 +963,9 @@ - (void)collectionViewCellWillAppear:(HUBComponentCollectionViewCell *)cell
     }
     
     [self componentWrapperWillAppear:wrapper];
-    [self.delegate viewController:self componentWithModel:wrapper.model willAppearInView:cell];
+
+    UIView * const component = HUBComponentLoadViewIfNeeded(cell.component);
```
JohnSundell: So `nil` check this here first. Also `componentView` or something would be a better name, since it's a `UIView`, not a component 🙂

tests/HUBViewControllerTests.m
```diff
@@ -1863,6 +1864,7 @@ - (void)viewController:(UIViewController<HUBViewController> *)viewController
       willAppearInView:(UIView *)componentView
 {
     XCTAssertEqual(viewController, self.viewController);
+    XCTAssertFalse([componentView isKindOfClass:[HUBComponentCollectionViewCell class]]);
```
JohnSundell: Maybe add an assertion to make sure you get the correct view back, not only "not a collection view cell"?

sources/HUBUtilities.h
krris: Right, we don't need this. I was sending bad object to this function and it led to following error
> Implicit conversion from nullable pointer to non-nullable pointer 
But now worries, I know how to fixed this 😉

krris:
@JohnSundell 🎾

PR #91, opened 2016-10-20
Don’t notify delegate twice when performing programmatic selection

JohnSundell:
This patch fixes a bug that would cause a HUBViewController’s delegate to be notified of a selection event twice when that selection was triggered programmatically (compared to when the user selects a component through the UI).

The reason the tests weren’t catching this was because the UI wasn’t setup properly before triggering the programmatic selection. This patch also fixes that test.


JohnSundell:
@spotify/objc-dev

@cerihughes This is the bug we encountered when writing TA tests today, when the component would be selected twice.

PR #94, opened 2016-10-22
Component wrapper: fix gesture recognizer stealing button touches

FranGamb:
The gesture recognizer used to handle component highlighting and selection prevents buttons inside the component from receiving touches. This PR fixes the issue by not forwarding button touches to the gesture recognizer.

### DIFF
--- a/sources/HUBComponentWrapper.m
+++ b/sources/HUBComponentWrapper.m
@@ -359,6 +359,15 @@ NS_ASSUME_NONNULL_BEGIN
     return YES;
 }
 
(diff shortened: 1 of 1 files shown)


sources/HUBComponentWrapper.m
```diff
@@ -359,6 +359,15 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecogni
     return YES;
 }
 
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceiveTouch:(UITouch *)touch
+{
+    if ([touch.view isKindOfClass:[UIButton class]]) {
```
JohnSundell: You can just return the negative result of this expression, turning this method into a one-liner 😄

PR #95, opened 2016-10-24
Fix bug causing nested components to be unelectable

JohnSundell:
Since the new selection handling mechanism was introduced, child components have been made unselectable, because the touch recognizer used to perform selection handling was “stealing touches” from nested views.

This has now been fixed by not using selection handling if one of the following view classes were touched:

- UIButton
- UICollectionViewCell
- UITableViewCell

A more rigid implementation of this would be to check the visible child components and hit test them, but since we’re currently also using plain UIViews as child components, we need to go through the view hierarchy.


JohnSundell:
@spotify/objc-dev

PR #98, opened 2016-10-24
Power up actions API to support content operations & async, chainable actions

JohnSundell:
This PR makes some significant upgrades to the Action API.

- Enable content operations to **perform** actions. They were previously able to **observe** actions, but now they can also perform them.
- Enable actions to be async (through the new `HUBAsyncAction` protocol). Async actions are retained until they have completed.
- Enable action chaining. Actions can now hand over to another one, and also supply a new set of custom data. This enables the construction of really nice user flows using actions, and enables API users to easily extend the framework.

To demonstrate these new capabilities, a "Todo list" feature has been added to the demo app, that enables the user to add new items to a todo list through an alert dialog (all controlled using async actions).

The Action Programming guide, as well as the Xcode templates, will be updated in a subsequent PR.

simulator screen shot 24 oct 2016 19 26 02
simulator screen shot 24 oct 2016 19 26 11

### DIFF
--- a/demo/sources/TodoListAddAction.swift
+++ b/demo/sources/TodoListAddAction.swift
@@ -1,0 +1,37 @@
+import Foundation
+import HubFramework
+
+/// Action custom data keys used by `TodoListAddAction`
+struct TodoListAddActionCustomDataKeys {
+    /// The title of the item to add to a todo list
+    static var itemTitle: String { return "item" }
+}
+
+/// Action that presents an alert to add a todo list item
+class TodoListAddAction: NSObject, HUBAsyncAction {
+    weak var delegate: HUBAsyncActionDelegate?
+    
+    func perform(with context: HUBActionContext) -> Bool {
+        let alertController = UIAlertController(title: "Add an item", message: nil, preferredStyle: .alert)
+        alertController.addTextField(configurationHandler: nil)
+        
+        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel) { _ in
+            alertController.dismiss(animated: true, completion: nil)
+            self.delegate?.actionDidFinish(self, chainToActionWithIdentifier: nil, customData: nil)
+        }
+        
+        let doneAction = UIAlertAction(title: "Add", style: .default) { _ in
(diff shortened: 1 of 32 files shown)


JohnSundell:
@spotify/objc-dev

sources/HUBViewControllerImplementation.m
```diff
@@ -140,8 +141,9 @@ - (instancetype)initWithViewURI:(NSURL *)viewURI
     _childComponentReusePool = [[HUBComponentReusePool alloc] initWithComponentRegistry:_componentRegistry
                                                                          UIStateManager:_componentUIStateManager];
     
-    _viewModelLoader.delegate = self;
-    _imageLoader.delegate = self;
+    viewModelLoader.delegate = self;
```
krris: Shouldn't it be `_viewModelLoader` instead `viewModelLoader`? Same question about 145 and 146 lines.

demo/sources/TodoListAddAction.swift
```diff
@@ -0,0 +1,37 @@
...
+        
+        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel) { _ in
+            alertController.dismiss(animated: true, completion: nil)
+            self.delegate?.actionDidFinish(self, chainToActionWithIdentifier: nil, customData: nil)
+        }
+        
+        let doneAction = UIAlertAction(title: "Add", style: .default) { _ in
+            let nextActionIdentifier = HUBIdentifier(namespace: TodoListActionFactory.namespace, name: TodoListActionNames.addCompleted)
+            let nextActionCustomData = [TodoListAddActionCustomDataKeys.itemTitle: alertController.textFields!.first!.text]
+            self.delegate?.actionDidFinish(self, chainToActionWithIdentifier: nextActionIdentifier, customData: nextActionCustomData)
```
krris: What do you think about naming it `actionIdentifier` instead of `chainToActionWithIdentifier`?

sources/HUBViewControllerImplementation.m
JohnSundell: It doesn't really matter, since they refer to the same object :)

demo/sources/TodoListAddAction.swift
JohnSundell: Because otherwise it's unclear that you are performing chaining.

PR #100, opened 2016-10-25
Calculate parent component’s view size up-front when creating children

JohnSundell:
This patch fixes a bug where child components would get a `CGSizeZero` container view size, if created before their parent component was resized by `UICollectionView`. This is done by using layout attributes directly when computing the container view size for root body components.

This lookup is very fast (`O(1)`), since we cache the layout attributes by index path.


JohnSundell:
@spotify/objc-dev

PR #103, opened 2016-10-25
Have 🚫Danger report build problems

rastersize:
This PR makes [🚫Danger](https://danger.systems) report back about build problems for each PR. It will report any build (compile and link time) warnings and error. This is in preparation for using Travis-CI Pro where external contributors might be able to read the build log themselves. When reported back it should look like: https://github.com/diogot/danger-xcode_summary#how-does-it-look

Two Ruby gems are brought in for this:
- `danger-xcode_summary` (reads and report back on Xcode build result).
- `xcpretty-json-formatter` (generates a JSON file that `danger-xcode_summary` can read).

Also bumps the CI tools Danger and Fastlane. The latter for improved Xcode 8 handling.

Additionally a Danger check for changes to the `LICENSE` file has been added. Any PR modifying it will be failed.

@spotify/objc-dev


rastersize:
And it also worked as evident by https://github.com/spotify/HubFramework/pull/103#issuecomment-256210564 😄 

Merging away.

PR #104, opened 2016-10-26
Diff updated indexes using old view model

Goos:
Changes so that the HUBViewModelDiff calculates updated indices using the previous data model rather than the new one.

@spotify/objc-dev (@cerihughes, @JohnSundell @rastersize)

### DIFF
--- a/tests/HUBCollectionViewLayoutTests.m
+++ b/tests/HUBCollectionViewLayoutTests.m
@@ -315,7 +315,7 @@
 - (void)testProposedContentOffsetAfterRecomputing
 {
     for (NSUInteger i = 0; i < 30; i++) {
-        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier];
+        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier preferredIndex:i];
     }
 
     id<HUBViewModel> const viewModel = [self.viewModelBuilder build];
@@ -339,17 +339,17 @@
     NSRange const addedRange = NSMakeRange(0, 2);
     NSIndexSet * const addedIndices = [NSIndexSet indexSetWithIndexesInRange:addedRange];
 
+    [addedIndices enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
+        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier preferredIndex:idx];
+        if (idx < currentIndex) {
+            insertionHeight += self.fullWidthComponent.preferredViewSize.height;
+        }
+    }];
+
     [removedIndices enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
         [self removeBodyComponentAtIndex:idx];
         if (idx <= currentIndex) {
             deletionHeight += self.fullWidthComponent.preferredViewSize.height;
-        }
-    }];
-
-    [addedIndices enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
-        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier preferredIndex:idx];
-        if (idx <= currentIndex) {
-            insertionHeight += self.fullWidthComponent.preferredViewSize.height;
         }
     }];
 
@@ -383,7 +383,7 @@
     [layout computeForCollectionViewSize:self.collectionViewSize viewModel:firstViewModel diff:nil];
     
     for (NSUInteger i = 0; i < 10; i++) {
-        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier];
+        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier preferredIndex:i];
     }
 
     id<HUBViewModel> const secondViewModel = [self.viewModelBuilder build];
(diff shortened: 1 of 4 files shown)


tests/HUBViewModelDiffTests.m
```diff
@@ -184,10 +184,68 @@ - (void)testComplexChangeSet
...
+    XCTAssert([diff.reloadedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:8 inSection:0]]);
     XCTAssert(diff.reloadedBodyComponentIndexPaths.count == 2);
     XCTAssert(diff.insertedBodyComponentIndexPaths.count == 2);
     XCTAssert(diff.deletedBodyComponentIndexPaths.count == 2);
 }
 
+- (void)testInsertionOfComponentModelAtStartWithDataChanges
+{
+    NSArray<id<HUBComponentModel>> *firstComponents = @[
+                                                        [self createComponentModelWithIdentifier:@"component-1" customData:nil],
```
JohnSundell: Looks like this code formatting is a bit off 🎨

tests/HUBViewModelDiffTests.m
```diff
@@ -184,10 +184,68 @@ - (void)testComplexChangeSet
     XCTAssert([diff.insertedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:2 inSection:0]]);
     XCTAssert([diff.insertedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:9 inSection:0]]);
     XCTAssert([diff.reloadedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:1 inSection:0]]);
-    XCTAssert([diff.reloadedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:7 inSection:0]]);
+    XCTAssert([diff.reloadedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:8 inSection:0]]);
```
JohnSundell: This seems a bit dangerous, or was the test incorrect?

tests/HUBViewModelDiffTests.m
JohnSundell: (also the same for the other array literals)

tests/HUBViewModelDiffTests.m
```diff
@@ -184,10 +184,68 @@ - (void)testComplexChangeSet
...
+    XCTAssert(diff.reloadedBodyComponentIndexPaths.count == 3);
+    XCTAssert(diff.insertedBodyComponentIndexPaths.count == 1);
+    XCTAssert(diff.deletedBodyComponentIndexPaths.count == 0);
+    XCTAssert([diff.insertedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:0 inSection:0]]);
+    XCTAssert([diff.reloadedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:0 inSection:0]]);
+    XCTAssert([diff.reloadedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:1 inSection:0]]);
+    XCTAssert([diff.reloadedBodyComponentIndexPaths containsObject:[NSIndexPath indexPathForItem:2 inSection:0]]);
+}
+
+- (void)testInsertionOfComponentModelsAtStartWithDataChanges
```
JohnSundell: Suggestion: I had to spend quite a few minutes to figure out what the difference was between this test and the one above. Renaming it to `testInsertionOfMultipleComponentModelsAtStartWithDataChanges` would make it a bit more explicit (IMO).

cerihughes: Sorry...  they were tests I put together quickly.

tests/HUBViewModelDiffTests.m
cerihughes: Isn't that the default Xcode formatter settings? (it is on my setup at least)

cerihughes:
@JohnSundell This should be used in preference to what I was working on (but we've pulled in the tests I wrote that reproduced the problem). Mine was a quick patch on top and needed an extra iteration, but this fixes the problem in the same number of iterations.

tests/HUBViewModelDiffTests.m
cerihughes: The implementation we had made some incorrect assumptions about the index paths you need to pass to `UICollectionView performBatchUpdates:completion:` so would sometimes generate incorrect index paths that caused the crash.

The new code generates correct index paths, so this test has to be updated to reflect the correct logic.

cerihughes:
Looks great 🎉

tests/HUBViewModelDiffTests.m
JohnSundell: Yeah it is, but our allmighty [code style](https://github.com/spotify/ios-style) says differently 😄

JohnSundell:
@cerihughes Ah got it, awesome! 👍

JohnSundell:
@Goos @cerihughes Really awesome work guys! ⭐️ ⭐️ ⭐️ ⭐️ ⭐️

tests/HUBViewModelDiffTests.m
JohnSundell: Ah makes sense 👍

Goos: This is due to the change – the updated item was at index 8 prior to the update :smile:

Goos:
@goos, @cerihughes Updated the tests and fixed the style issues 😄

tests/HUBCollectionViewLayoutTests.m
```diff
@@ -339,17 +339,17 @@ - (void)testProposedContentOffsetAfterRecomputing
...
+    [addedIndices enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
+        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier preferredIndex:idx];
+        if (idx < currentIndex) {
+            insertionHeight += self.fullWidthComponent.preferredViewSize.height;
         }
     }];
 
-    [addedIndices enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
-        [self addBodyComponentWithIdentifier:self.fullWidthComponentIdentifier preferredIndex:idx];
+    [removedIndices enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
```
JohnSundell: Could you just give some quick context as to why these tests had to be changed? 😄

Goos: It was actually a problem with the test that I hadn't seen before :smile: It wasn't adding & removing the intended components

JohnSundell: Ah alright - thanks 🌮

PR #107, opened 2016-10-26
View controller: Add property to inspect whether view is scrolling

JohnSundell:
This will enable API users to make decisions based on whether the user is currently scrolling the main content view of a Hub Framework view controller.


JohnSundell:
@spotify/objc-dev

PR #109, opened 2016-10-26
View controller: Take content offset into account for hit testing

JohnSundell:
This patch fixes a bug in the hit testing API `indexOfBodyComponentAtPoint:`, which previously wouldn’t take the content offset of the view into account.

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -326,6 +326,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSUInteger)indexOfBodyComponentAtPoint:(CGPoint)point
 {
+    point.y += self.collectionView.contentOffset.y;
+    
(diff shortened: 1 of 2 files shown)


JohnSundell:
@spotify/objc-dev

Goos:
Wouldn't it make more sense if it expected a point in the scrollView's coordinate coordinate space though? Seeing as that's what `frameForBodyComponentAtIndex:` returns (and I know we're using it like that in a scrollhandler in a certain app :smile:).

JohnSundell:
@Goos The thing is that the scroll view is not exposed as part of the API, so then we'd have to add an extra step where you convert your point from the view controller's coordinate system to the scroll view's internal coordinate system. Feels like an extra step. The use case that I've seen so far (using this in a gesture recognizer to hit test components) is translating from a container view controller's coordinate space into a component - which is nice if the framework handles for you 😄

PR #110, opened 2016-10-27
Don’t compare index path for equality.

cerihughes:
The index path shouldn't be used when considering equality as it uses information about the parent, (which is already excluded from the calculation).


cerihughes:
@Goos @JohnSundell as discussed yesterday.

PR #113, opened 2016-10-27
View controller: Add API to retrieve a single visible component view

JohnSundell:
This new API is an alternative to returning **all** visible component views, and provides a much faster (pretty much `O(1)` complexity - if you disregard the length of the index path) way of looking up just a **single** component view.

### DIFF
--- a/include/HubFramework/HUBViewController.h
+++ b/include/HubFramework/HUBViewController.h
@@ -168,15 +168,29 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Returns the views of the components of the given type that are currently visible on screen, keyed by their index path
  *
+ *  @param componentType The type of component to check for visiblilty.
+ *
  *  The index paths used for keys contains the indexes for the components' views, starting from the root. For example,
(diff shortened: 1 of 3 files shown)


JohnSundell:
@spotify/objc-dev

cerihughes:
Sorry for the delay in reviewing, but why doesn't `visibleViewForComponentOfType:indexPath:` just do

```
- (nullable UIView *)visibleViewForComponentOfType:(HUBComponentType)componentType indexPath:(NSIndexPath *)indexPath
{
    return [self visibleComponentViewsForComponentType:componentType][indexPath];
}
```

cerihughes:
Oh, for complexity reasons?

cerihughes:
Sorry. I should have read the PR description!

cerihughes:
👍

JohnSundell:
@cerihughes Yeah exactly, basically in the Spotify app, we're calling this as part of handling a gesture recognizer, so need to have close to `O(1)` complexity to not hog the main tread 😄

PR #114, opened 2016-10-27
Moves the configuration of header and overlay components when reloading

Goos:
This caused some issues when attempting to scroll somewhere in the initial `viewControllerDidFinishRendering:`-call, as `configureHeaderComponent` would change the content offset _after_ the delegate was notified (as the completion block is called instantly).

@spotify/objc-dev


JohnSundell:
@Goos You think it's worth adding a test that makes sure this doesn't happen again? :)

Goos:
@JohnSundell Sure 😛

Goos:
@JohnSundell Added a test case – mind taking another look? :smile:

Issue #29, opened 2016-09-29
Open source live editing tools


JohnSundell:
Internally, we have a command line tool written in Swift that enables you to live edit a Hub Framework-powered UI by editing a local JSON file. The tool is currently pretty hacky and only works with the Spotify app, but we should clean it up and include it in this repo so that anyone can use it!

PR #118, opened 2016-10-28
Introduce Live Editing

JohnSundell:
This PR adds live editing capabilites to the Hub Framework. 🎉 This is done through a new `HubFrameworkLive` project that builds a `hublive` command line application, that can then connect to a newly introduced `HUBLiveService` running in a Hub Framework powered application.

**A few notes:**

- The implementation of the `HUBLiveService` API is only compiled if the host application is being compiled for `DEBUG`. This reduces the risk of running into security issues in production code.
- The live editing system is built using `NSNetService` - which is the Apple-provided framework for interacting with the Bonjour local networking system. It uses a socket on a user configurable port.
- Using live is opt-in for applications using the Hub Framework. The live service needs to be explicitly enabled by calling `hubManager.liveService?.start(onPort: 5555)`.

**Usage:**

After installing `hublive`, simply start the live service (the demo app already does this automatically). You can then call `hublive /path/to/json/file.json`, and live edit away 🚀

Resolves https://github.com/spotify/HubFramework/issues/29

liveediting

### DIFF
--- a/sources/HUBLiveContentOperation.h
+++ b/sources/HUBLiveContentOperation.h
@@ -1,0 +1,46 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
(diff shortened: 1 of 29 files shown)


sources/HUBLiveContentOperation.h
```diff
@@ -0,0 +1,46 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#ifdef DEBUG
```
8W9aG: Shouldn't this be if DEBUG rather than ifdef? DEBUG should be 1 presumably. Although perhaps make a different flag that gets set when DEBUG = 1?

8W9aG: If that is the case shouldn't we do ifdef then if rather ifdef only? What if DEBUG=0, doesn't that mean DEBUG isn't on semantically?

8W9aG: Better yet create a header where ifndef DEBUG then define it to 0?

JohnSundell:
@8W9aG Fixed the macro issue and added unit tests + CI. Wanna take a 2nd look? 😄

8W9aG:
👍

PR #120, opened 2016-10-30
Live CLI: Improve error handling

JohnSundell:
- The program will send data async if a Socket wasn’t immediately connected.
- More thorough error messages.

### DIFF
--- a/live/sources/SocketClient.swift
+++ b/live/sources/SocketClient.swift
@@ -64,8 +64,28 @@ class SocketClient {
             return
         }
         
+        switch stream.streamStatus {
+        case .notOpen:
+            printError(message: "Stream is not open")
+        case .opening, .writing:
+            DispatchQueue(label: "Awaiting stream setup").async {
+                self.send(data: data)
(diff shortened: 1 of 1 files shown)


JohnSundell:
@spotify/objc-dev

live/sources/SocketClient.swift
```diff
@@ -64,16 +64,36 @@ class SocketClient {
             return
         }
         
+        switch stream.streamStatus {
+        case .notOpen:
+            printError(message: "Stream is not open")
+        case .opening, .writing:
+            DispatchQueue(label: "Awaiting stream setup").async {
```
Goos: Wouldn't it be better to create a dedicated queue when initializing or using the global queue rather than creating a new one each write?

JohnSundell: Sure, I can do that. Can lazily initialize it when needed. Good idea 👍

Goos: Awesome 👌

Issue #28, opened 2016-09-29
Add pagination support


JohnSundell:
`HUBViewModel` has an `extensionURL` property, which is not used yet. The idea is that once the user scrolls to the bottom of the view, more content should be loaded automatically from that URL (probably by triggering a new API on `HUBContentOperation`, or an extended protocol) and appended to the current view model.

It should work the same way as a "See all" view does in the Search feature in the Spotify app.

JohnSundell:
Starting work on this now. We will remove `extensionURL` in favor of a pagination API for content operations. This has also been synced with Android - they are not using `extensionURL` either.

PR #121, opened 2016-10-31
Remove HUBViewModel.extensionURL

JohnSundell:
This property was originally intended to provide a way to paginate content from a backend service, by letting that backend service returning a URL from where additional content can be loaded.

However, since the move to content operations (and with content operation chaining) - this has been made less and less relevant, since content operations are not required to actually load their content from a backend service.

This property wasn’t actually used anywhere in the framework, so removing it in favor of an upcoming new content operation extension API for paginated content.

Relates to https://github.com/spotify/HubFramework/issues/28


JohnSundell:
@spotify/objc-dev

PR #122, opened 2016-11-01
Demo app: Make image components user intractable

JohnSundell:
`UIImageView` is `userInteractionEnabled = false` per default.


JohnSundell:
@spotify/objc-dev

JohnSundell:
@goos You navigate 🚀

Issue #70, opened 2016-10-12
Remove NSObject inheritance from all protocols.


krris:
Currently each protocol in HubFramework looks like this:

```
@protocol HUBImageLoader <NSObject>
- (void)foo;
@end
```

IMO we can remove inheritance from `NSObject` in each protocol as far as we don't use any `NSObject` methods ([here](http://stackoverflow.com/questions/679822/why-tack-a-protocol-of-nsobject-to-a-protocol-implementation) is a discussion about this if you want to read more). What does it give us?
- Our code will be more _swifty_. 
- While writing in Swift we don't need to inherit from NSObject.
- Less code in our codebase 😉 

What do you think about this?

JohnSundell:
I'm all for it if it doesn't have any side effects! @rastersize what do you think?

rastersize:
AFAIK there are really only two side effects. The first is that we can’t use `@optional` to mark methods as optional in protocol. Fortunately we don’t do that 😊 Secondly we’d technically not know if our objects respond to a bunch of methods we might need, such as `-isEqual:`.

But it would be great to get rid of the conformance. Something for [2.1](https://github.com/spotify/HubFramework/milestone/1) I’d say. The tasks we need to complete to get to this state are at least:
- Audit every place where we assume an object is of type `NSObject`.
  - There are a few places where we type things as `NSObject<Something>`. This is probably not always true at the moment anyhow and will likely not be true afterwards.
- Make sure anything that _needs_, e.g., `-isEqual:` continues to conform to `NSObject` or something similar.
- Add linter making sure we don’t introduce new protocols conforming to `NSObject`.
- Add linter making sure we don’t introduce `@optional` in a protocol.
- Remove conformance of `NSObject` from our protocols.

All in all, I’m in favor of this proposal.

JohnSundell:
Ok, let's consider this proposal passed! 👍 @krris want to proceed with the implementation?

krris:
@JohnSundell I would love to.

krris:
What about using http://oclint.org/ as a linter? 👍  or 👎  ?

JohnSundell:
I'm not opposed to trying OCLint out. Seems to be widely used as well. @rastersize, @8W9aG, @cerihughes, any opinions?

rastersize:
@krris @JohnSundell The problem we’ve encountered with linters so far is the pain of getting them configured for our code style. But if we can get that to work I’m all for linting.

rastersize:
[...]

PR #123, opened 2016-11-01
Adds possibility of scrolling to indexpaths (children included) to the HUBViewController.

Goos:
Putting this up as work-in-progress for discussion purposes! 

This PR adds the possibility of scrolling to nested children. 

The first challenge with doing this is that the HubFramework doesn't really know anything about how its children handles _their_ children, which forces us to delegate that to the components themselves. 

The second challenge is animations – since the components are reported to the HubFramework when they are visible, we have to wait for the animations to finish each step along the way.

@JohnSundell, @cerihughes, @spotify/objc-dev

### DIFF
--- a/include/HubFramework/HUBComponentWithChildren.h
+++ b/include/HubFramework/HUBComponentWithChildren.h
@@ -110,6 +110,19 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, weak, nullable) id<HUBComponentChildDelegate> childDelegate;
 
+/**
+ * Called when programmatically scrolling to a child within this parent component.
+ *
+ * @param childIndex The index of the component that is being scrolled to.
+ * @param scrollPosition The preferred position of the component after scrolling.
+ * @param animated Whether or not the scrolling should be animated.
+ * @param completionHandler The block to call once the component is visible.
+ */
+- (void)scrollToComponentAtIndex:(NSUInteger)childIndex
+                atScrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                        animated:(BOOL)animated
+                      completion:(void (^)())completionHandler;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/include/HubFramework/HUBComponentWithChildren.h
+++ b/include/HubFramework/HUBComponentWithChildren.h
@@ -110,19 +110,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, weak, nullable) id<HUBComponentChildDelegate> childDelegate;
 
-/**
- * Called when programmatically scrolling to a child within this parent component.
- *
- * @param childIndex The index of the component that is being scrolled to.
- * @param scrollPosition The preferred position of the component after scrolling.
- * @param animated Whether or not the scrolling should be animated.
- * @param completionHandler The block to call once the component is visible.
- */
-- (void)scrollToComponentAtIndex:(NSUInteger)childIndex
-                atScrollPosition:(UICollectionViewScrollPosition)scrollPosition
-                        animated:(BOOL)animated
-                      completion:(void (^)())completionHandler;
-
 @end
 
 NS_ASSUME_NONNULL_END
--- a/include/HubFramework/HUBComponentWithScrolling.h
+++ b/include/HubFramework/HUBComponentWithScrolling.h
@@ -1,0 +1,43 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import "HUBComponentWithChildren.h"
+
+/**
+ * Extended Hub component protocol that adds the ability scroll between child components.
+ *
+ * Use this protocol if your component supports scrolling between components within it. 
+ * See `HUBComponent` for more info.
+ */
+@protocol HUBComponentWithScrolling <HUBComponentWithChildren>
+
+/**
+ * Called when programmatically scrolling to a child within this parent component.
+ *
+ * @param childIndex The index of the component that is being scrolled to.
+ * @param animated Whether or not the scrolling should be animated.
+ * @param completionHandler The block to call once the component is visible.
+ */
+- (void)scrollToComponentAtIndex:(NSUInteger)childIndex
+                        animated:(BOOL)animated
+                      completion:(void (^)())completionHandler;
+
+@end
--- a/include/HubFramework/HUBComponentWithScrolling.h
+++ b/include/HubFramework/HUBComponentWithScrolling.h
@@ -33,10 +33,12 @@
  * Called when programmatically scrolling to a child within this parent component.
  *
  * @param childIndex The index of the component that is being scrolled to.
+ * @param scrollPosition The preferred position of the component after scrolling.
  * @param animated Whether or not the scrolling should be animated.
  * @param completionHandler The block to call once the component is visible.
  */
 - (void)scrollToComponentAtIndex:(NSUInteger)childIndex
+                  scrollPosition:(UICollectionViewScrollPosition)scrollPosition
                         animated:(BOOL)animated
                       completion:(void (^)())completionHandler;
 
--- a/include/HubFramework/HUBComponentWithScrolling.h
+++ b/include/HubFramework/HUBComponentWithScrolling.h
@@ -23,7 +23,7 @@
 #import "HUBScrollPosition.h"
 
 /**
- * Extended Hub component protocol that adds the ability scroll between child components.
+ * Extended Hub component protocol that adds the ability to scroll between child components.
  *
  * Use this protocol if your component supports scrolling between components within it. 
  * See `HUBComponent` for more info.
--- a/include/HubFramework/HUBComponentWithScrolling.h
+++ b/include/HubFramework/HUBComponentWithScrolling.h
@@ -20,6 +20,7 @@
  */
 
 #import "HUBComponentWithChildren.h"
+#import "HUBScrollPosition.h"
 
 /**
  * Extended Hub component protocol that adds the ability scroll between child components.
@@ -35,11 +36,11 @@
  * @param childIndex The index of the component that is being scrolled to.
  * @param scrollPosition The preferred position of the component after scrolling.
  * @param animated Whether or not the scrolling should be animated.
- * @param completionHandler The block to call once the component is visible.
+ * @param completion The block to call once the component is visible.
  */
 - (void)scrollToComponentAtIndex:(NSUInteger)childIndex
-                  scrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                  scrollPosition:(HUBScrollPosition)scrollPosition
                         animated:(BOOL)animated
-                      completion:(void (^)())completionHandler;
+                      completion:(void (^)(void))completion;
 
 @end
--- a/include/HubFramework/HUBScrollPosition.h
+++ b/include/HubFramework/HUBScrollPosition.h
@@ -1,0 +1,36 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+/**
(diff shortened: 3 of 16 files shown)


Goos:
Not entirely sure if we should keep the `scrollPosition` parameter to the HUBComponentWithChildren-call or just let them decide where to put it themselves

include/HubFramework/HUBComponentWithChildren.h
```diff
@@ -110,6 +110,21 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ * Called when programmatically scrolling to a child within this parent component.
+ *
+ * @param childIndex The index of the component that is being scrolled to.
+ * @param scrollPosition The preferred position of the component after scrolling.
+ * @param animated Whether or not the scrolling should be animated.
+ * @param completionHandler The block to call once the component is visible.
+ */
+- (void)scrollToComponentAtIndex:(NSUInteger)childIndex
+                atScrollPosition:(UICollectionViewScrollPosition)scrollPosition
```
Goos: Not entirely sure if we should keep this parameter or just let them decide where to put it themselves

include/HubFramework/HUBComponentWithChildren.h
```diff
@@ -110,6 +110,21 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, weak, nullable) id<HUBComponentChildDelegate> childDelegate;
 
+@optional
```
JohnSundell: Don't use optional methods 😱 They're not compile time safe and make the API loose its explicitness. In fact, as part of https://github.com/spotify/HubFramework/issues/70 we will start linting for `@optional` and fail the build if they're used. Instead, either make it a required method (if it should be strongly coupled with this API), or make a new protocol that extends `HUBComponent`.

Goos:
@cerihughes, @JohnSundell do you think it makes sense to call the `scrollingWillStartInViewController:` & `scrollingDidEndInViewController:` methods on the scroll handler when manually scrolling to a component indexpath?

JohnSundell:
@Goos It probably does, because we can't make the assumption that the scrolling is both performed and observed in the same piece of code. That is, some logic in an app might cause the view to scroll, while the scroll handler for the current view is implemented in a totally different code base.

sources/HUBComponentWrapper.m
```diff
@@ -307,6 +314,21 @@ - (void)updateViewForSelectionState:(HUBComponentSelectionState)selectionState
...
+
+- (void)scrollToComponentAtIndex:(NSUInteger)childIndex
+                  scrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                        animated:(BOOL)animated
+                      completion:(void (^)())completionHandler
+{
+    if ([self.component conformsToProtocol:@protocol(HUBComponentWithScrolling)]) {
+        [(id<HUBComponentWithScrolling>)self.component scrollToComponentAtIndex:childIndex
+                                                                 scrollPosition:scrollPosition
+                                                                      animated:animated
```
ShervinS: Indentation? :D

sources/HUBViewControllerImplementation.m
```diff
@@ -1314,6 +1337,103 @@ - (void)removeComponentWrapperFromLookupTables:(nullable HUBComponentWrapper *)c
     }
 }
 
+- (void)scrollToRootComponentAtIndex:(NSUInteger)componentIndex
+                      scrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                            animated:(BOOL)animated
+                          completion:(void (^)())completionHandler
+{
+    NSUInteger const componentCount = (NSUInteger)[self.collectionView numberOfItemsInSection:0];
+    if (componentIndex == NSNotFound || componentIndex >= componentCount) {
```
Goos: @JohnSundell – do we normally assert these things or fail silently?

JohnSundell: Normally, we assert - since it's an incorrect use of the API. Failing silently has the downside that no-one knows that they did something wrong, until we get a bug report 😄

Goos:
@JohnSundell, @cerihughes unit tests are in place – would you mind taking another look? :smile:

include/HubFramework/HUBComponentWithScrolling.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Called when programmatically scrolling to a child within this parent component.
+ *
+ * @param childIndex The index of the component that is being scrolled to.
+ * @param scrollPosition The preferred position of the component after scrolling.
+ * @param animated Whether or not the scrolling should be animated.
+ * @param completionHandler The block to call once the component is visible.
+ */
+- (void)scrollToComponentAtIndex:(NSUInteger)childIndex
+                  scrollPosition:(UICollectionViewScrollPosition)scrollPosition
```
JohnSundell: With this, we're kind of assuming that the component will be implemented using `UICollectionView`, which is not always true. Do we need to support multiple scroll positions at this point? If so - I'd say make a custom `enum` that is not tied to `UICollectionView`.

include/HubFramework/HUBComponentWithScrolling.h
```diff
@@ -0,0 +1,45 @@
...
+                        animated:(BOOL)animated
+                      completion:(void (^)())completionHandler;
```
JohnSundell: Make the external parameter name match the documentation and the internal name - `completionHandler`.

include/HubFramework/HUBViewController.h
```diff
@@ -171,6 +171,19 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)scrollToContentOffset:(CGPoint)contentOffset animated:(BOOL)animated;
 
 /**
+ *  Scroll to a desired content offset.
```
JohnSundell: Not content offset, right? To a component at a given index path?

include/HubFramework/HUBViewController.h
```diff
@@ -171,6 +171,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ *  @param componentIndexPath The index path of the component to scroll to.
+ *  @param scrollPosition The preferred position of the component after scrolling.
+ *  @param animated Whether or not the scrolling should be animated.
+ *  @param completionHandler A block that is called once the component at the provided index path is visible.
+ */
+- (void)scrollToComponentAtIndexPath:(NSIndexPath *)componentIndexPath
+                      scrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                            animated:(BOOL)animated
+                          completion:(void (^ _Nullable)())completionHandler;
```
JohnSundell: Same here - `completionHandler`

include/HubFramework/HUBViewControllerScrollHandler.h
```diff
@@ -110,4 +110,15 @@
...
                                        proposedContentOffset:(CGPoint)proposedContentOffset;
 
+/**
+ * Return the content offset for displaying a component at a certain scroll position.
+ * 
+ * @param componentIndex The index of the component to display.
+ * @param scrollPosition The position to display the component at.
+ * @param viewController The view controller in question.
+ */
+- (CGPoint)contentOffsetForDisplayingComponentAtIndex:(NSUInteger)componentIndex
```
JohnSundell: Hmmm, why is this needed? 🤔 Should the scroll handler have an opinion on where the children are?

sources/HUBComponentWrapper.h
```diff
@@ -196,6 +198,12 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState;
  */
 - (void)saveComponentUIState;
 
+/** 
+ *  Returns the child component wrapper located at the provided index – if visible. 
+ *  @param index The index of the component to retrieve.
```
JohnSundell: 🎨 New line between method description and `@param`

sources/HUBComponentWrapper.h
```diff
@@ -196,6 +198,12 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState;
...
+ */
+- (nullable HUBComponentWrapper *)childComponentAtIndex:(NSUInteger)index;
```
JohnSundell: Since this is only returning visible components - how about naming it `visibleChildComponentAtIndex:`?

sources/HUBComponentWrapper.m
```diff
@@ -118,6 +120,11 @@ - (void)saveComponentUIState
     }
 }
 
+- (nullable HUBComponentWrapper *)childComponentAtIndex:(NSUInteger)index
+{
+    return self.childrenByIndex[@(index)];
```
JohnSundell: This doesn't only contain visible children. It contains all children created. So the docs doesn't match 🙂

sources/HUBComponentWrapper.m
```diff
@@ -307,6 +314,21 @@ - (void)updateViewForSelectionState:(HUBComponentSelectionState)selectionState
...
 }
 
+#pragma mark - HUBComponentWithScrolling
+
+- (void)scrollToComponentAtIndex:(NSUInteger)childIndex
+                  scrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                        animated:(BOOL)animated
+                      completion:(void (^)())completionHandler
+{
+    if ([self.component conformsToProtocol:@protocol(HUBComponentWithScrolling)]) {
```
JohnSundell: 🎨 Return early instead of indenting method?

sources/HUBViewControllerDefaultScrollHandler.m
```diff
@@ -68,6 +69,26 @@ - (CGPoint)targetContentOffsetForEndedScrollInViewController:(UIViewController<H
     return proposedContentOffset;
 }
 
+- (CGPoint)contentOffsetForDisplayingComponentAtIndex:(NSUInteger)componentIndex
+                                       scrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                                       viewController:(UIViewController<HUBViewController> *)viewController
+{
+    CGRect const componentFrame = [viewController frameForBodyComponentAtIndex:componentIndex];
```
JohnSundell: Really cuirous as to why this should be the responsibility of the scroll handler. Seems like something the view controller should do.

sources/HUBViewControllerImplementation.m
```diff
@@ -1314,6 +1337,105 @@ - (void)removeComponentWrapperFromLookupTables:(nullable HUBComponentWrapper *)c
...
 }
 
+- (void)scrollToRootComponentAtIndex:(NSUInteger)componentIndex
+                      scrollPosition:(UICollectionViewScrollPosition)scrollPosition
+                            animated:(BOOL)animated
+                          completion:(void (^)())completionHandler
+{
+    NSUInteger const componentCount = (NSUInteger)[self.collectionView numberOfItemsInSection:0];
+
+    NSParameterAssert(componentIndex != NSNotFound);
```
JohnSundell: This is not needed as your next check will fail if the value is `NSNotFound` (since it's `INT_MAX`)

sources/HUBViewControllerImplementation.m
```diff
@@ -1314,6 +1337,105 @@ - (void)removeComponentWrapperFromLookupTables:(nullable HUBComponentWrapper *)c
...
+    if (indexPosition > 0) {
+        NSParameterAssert(childIndex <= componentWrapper.model.children.count);
+    }
+
+    __weak HUBViewControllerImplementation *weakSelf = self;
+    void (^stepCompletionHandler)() = ^{
+        HUBViewControllerImplementation *strongSelf = weakSelf;
+        // If the scrolling isn't animated we have to wait for a runloop cycle, as scrolling is deferred.
+        if (!animated) {
+            [[NSRunLoop currentRunLoop] runUntilDate:[NSDate date]];
```
JohnSundell: Won't this block the main thread for the remainder of the cycle? Wouldn't it be better to `dispatch_async()`?

include/HubFramework/HUBViewControllerScrollHandler.h
Goos: I figured it made sense that the scroll handler had a say in all kinds of scrolling (i.e. programmatic and interactive scrolling) :smile:

include/HubFramework/HUBComponentWithScrolling.h
tammofreese: How about making the block type more concrete by using `void (^)(void)` instead of `void (^)()`?

Goos:
@JohnSundell fixed a bunch of stuff – mind taking another look? :smile:

include/HubFramework/HUBViewController.h
```diff
@@ -171,6 +172,28 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)scrollToContentOffset:(CGPoint)contentOffset animated:(BOOL)animated;
 
 /**
+ *  Scroll to a desired component. Each index in the index path refers to one level of children.
+ *  For example, in order to scroll to a root component at an index, you would provide an index
+ *  path with that single index. If that component in turn has children, you can scroll between
+ *  those by providing an index path with two index, starting with the root index, and so on.
+ *
+ *  In order for child components to support nested scrolling, they must implement @c HUBComponentWithScrolling.
```
krris: What is `@c`?

include/HubFramework/HUBComponentWithScrolling.h
```diff
@@ -0,0 +1,46 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import "HUBComponentWithChildren.h"
+#import "HUBScrollPosition.h"
+
+/**
+ * Extended Hub component protocol that adds the ability scroll between child components.
```
krris: 🔬  ... ability **to** scroll

include/HubFramework/HUBScrollPosition.h
```diff
@@ -0,0 +1,36 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import <UIKit/UIKit.h>
+
+/**
+ * An enumeration to be used when scrolling to a specific position within a scrollable component.
+ * For ease of use, the different options map to their corresponding UICollectionViewScrollPosition.
```
JohnSundell: Smart! 👍

include/HubFramework/HUBViewController.h
```diff
@@ -171,6 +172,28 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ *  @param componentType The type of component you want to scroll to.
+ *  @param indexPath The index path of the component to scroll to.
+ *  @param scrollPosition The preferred position of the component after scrolling.
+ *  @param animated Whether or not the scrolling should be animated.
+ *  @param completion A block that is called once the component at the provided index path is visible.
+ *
+ *  @seealso HUBComponentWithScrolling
+ */
+- (void)scrollToComponentOfType:(HUBComponentType)componentType
```
JohnSundell: Awesome 🚀

include/HubFramework/HUBViewControllerScrollHandler.h
JohnSundell: If we don't need this API - why add it? 😄 It just makes the scroll handler more complicated. Why can't we use the normal content offset calculating method? 🙂

include/HubFramework/HUBViewController.h
Goos: It gives the following word a monospaced font :smile:

PR #124, opened 2016-11-02
customData support added to the child actions

rordulu:
customData can be sent for component actions, however there was no support to send customData when a child action was performed.

### DIFF
--- a/include/HubFramework/HUBComponentWithChildren.h
+++ b/include/HubFramework/HUBComponentWithChildren.h
@@ -84,11 +84,12 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  @param component The parent component
  *  @param childIndex The index of the child component that was selected
+ *  @param customData  Any custom data to use when the selection is handled. Will be available on the `HUBActionContext` passed to any actions handling the selection.
  *
  *  If your component has nested child components, you should call this method every time a child component was
  *  selected by the user, to enable the Hub Framework to handle the selection.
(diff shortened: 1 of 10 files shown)


include/HubFramework/HUBComponentWithChildren.h
```diff
@@ -84,11 +84,13 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  @param component The parent component
  *  @param childIndex The index of the child component that was selected
+ *  @param customData Any custom data to send to the action. Will be available on the
```
JohnSundell: The rest of the API doesn't mention anything about actions. So this documentation is a bit out of context. How about doing something like:

```
Any custom data to use when the selection is handled. Will be available on the `HUBActionContext` passed to any actions handling the selection.
```

include/HubFramework/HUBViewController.h
```diff
@@ -201,14 +201,16 @@ NS_ASSUME_NONNULL_BEGIN
  *  Perform a programmatic selection of a component with a given model
  *
  *  @param componentModel The model of the component to select
+ *  @param customData Any custom data to send to the action. Will be available on the
```
JohnSundell: Same comment here as for the documentation for `HUBComponentWithChildren`.

JohnSundell:
@rordulu The demo app needs to be updated with the new APIs in order for the build to succeed 🙂

PR #125, opened 2016-11-03
Testing: Add UI test

cerihughes:
This PR adds the ability to run UI tests against the demo app to make sure we don't have any regressions in our gesture recognisers and other touch input code.

At the moment, the test doesn't assert very much and isn't very useful. This PR is just setting up the required infrastructure.

There doesn't seem to be any special config in the `.travis.yml` file, so I'm hoping that the build "just works" and will pick up the new test target automatically!

### DIFF
--- a/demo/tests/HubFrameworkDemoUITests.swift
+++ b/demo/tests/HubFrameworkDemoUITests.swift
@@ -44,6 +44,9 @@ class HubFrameworkDemoUITests: XCTestCase {
         let collectionViewCell = collectionView.children(matching: .cell).element(boundBy: 1)
         let imageView = collectionViewCell.otherElements.children(matching: .image).element
         imageView.tap()
+
+        // Assert we've navigated away...
+        XCTAssertFalse(app.navigationBars["Pretty Pictures"].exists)
     }
 
     func testTappingNestedComponent() {
@@ -59,6 +62,9 @@ class HubFrameworkDemoUITests: XCTestCase {
         let nestedCollectionViewCell = collectionViewCell.cells.element(boundBy: 1)
         let imageView = nestedCollectionViewCell.otherElements.children(matching: .image).element
         imageView.tap()
+
+        // Assert we've navigated away...
+        XCTAssertFalse(app.navigationBars["Pretty Pictures"].exists)
     }
 
     /// This function walks the view hierarchy to find the hub framework's collection view.
--- a/demo/tests/HubFrameworkDemoUITests.swift
+++ b/demo/tests/HubFrameworkDemoUITests.swift
@@ -44,7 +44,15 @@ class HubFrameworkDemoUITests: XCTestCase {
         collectionView.cells.collectionViews.children(matching: .cell).element(boundBy: 0).otherElements.children(matching: .image).element.tap()
     }
 
+    /// This function walks the view hierarchy to find the hub framework's collection view.
+    /// There are currently no accessibility elements set on the collection view or its cells, so we have no other option.
+    /// - Parameter app: the top-level app instance.
+    /// - Returns: the collection view element.
     private func rootCollectionView(for app: XCUIApplication) -> XCUIElement {
-        return app.otherElements.containing(.navigationBar, identifier:"Pretty Pictures").children(matching: .other).element.children(matching: .other).element.children(matching: .other).element.children(matching: .collectionView).element
+        let navigatonBarParentQuery = app.otherElements.containing(.navigationBar, identifier:"Pretty Pictures")
+        let navigationTransitionView = navigatonBarParentQuery.children(matching: .other).element
+        let viewControllerWrapperView = navigationTransitionView.children(matching: .other).element
+        let hubContainerView = viewControllerWrapperView.children(matching: .other).element
+        return hubContainerView.children(matching: .collectionView).element
     }
 }
--- a/demo/tests/HubFrameworkDemoUITests.swift
+++ b/demo/tests/HubFrameworkDemoUITests.swift
@@ -44,7 +44,7 @@ class HubFrameworkDemoUITests: XCTestCase {
         collectionView.cells.collectionViews.children(matching: .cell).element(boundBy: 0).otherElements.children(matching: .image).element.tap()
     }
 
-    private func rootCollectionView(for app:XCUIApplication) -> XCUIElement {
+    private func rootCollectionView(for app: XCUIApplication) -> XCUIElement {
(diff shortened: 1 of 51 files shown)


fastlane/Fastfile
```diff
@@ -58,6 +58,40 @@ platform :ios do
...
 
+  desc 'Run ui tests'
+  lane :uitest do
+    system('pkill', '-9', '-x', 'Simulator')
+    UI.success('Simulator app killed')
+
+    device = ENV['TEST_DEVICE_NAME'] || 'iPhone 6s'
+    UI.message("Simulator testing device set to '#{device}'")
+
+    output_directory = 'build/tests'
```
rastersize: Change to `'build/ui-tests'` so we don’t overwrite the results of the unit tests.

fastlane/Fastfile
```diff
@@ -58,6 +58,40 @@ platform :ios do
   end
 
 
+  desc 'Run ui tests'
+  lane :uitest do
```
rastersize: This looks like an identical copy-paste of the `:test` lane. Except the value passed for the `:scheme` parameter. Could we instead make both of these use a shared lane that takes a `:scheme` parameter?

See https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Advanced.md#passing-parameters for more info on passing parameters to lanes.

---

Maybe something like:
```ruby
lane :test do
  run_tests(scheme: 'HubFramework', output_label: 'tests')
end

lane :test_ui do
  run_tests(scheme: 'HubFrameworkDemo', output_label: 'ui-tests')
end

private_lane :run_tests do |options|
  scheme = options[:scheme]
  output_label = options[:output_label]

  UI.user_error!("No Xcode scheme provided") unless scheme
  UI.user_error!("No label for the output was provided") unless output_label

  system('pkill', '-9', '-x', 'Simulator')
  UI.success('Simulator app killed')

  device = ENV['TEST_DEVICE_NAME'] || 'iPhone 6s'
  UI.message("Simulator testing device set to '#{device}'")

  output_directory = "build/#{output_label}"
  ENV['XCPRETTY_JSON_FILE_OUTPUT'] = "#{output_directory}/summary.json"

  scan(
    :workspace => 'HubFramework.xcworkspace',
    :scheme => scheme,
    :devices => [device],
    :derived_data_path => 'build/DerivedData',
    :output_directory => output_directory,
    :buildlog_path => 'build/logs/tests',
    :code_coverage => true,
    :output_types => '',
    :fail_build => false,
    :formatter => 'xcpretty-json-formatter'
  )

  trainer(
    path: 'build',
    output_directory: output_directory,
    extension: '.junit'
  )

  codecov() if ENV['CODECOV_TOKEN'] or ENV['TRAVIS'] == 'true'
end
```

rastersize:
We should also make sure to move the Codecov step to be ran after both of the test lanes have executed. Otherwise we’ll be processing and uploading the same result twice for (in this case) the unit tests.

demo/tests/HubFrameworkDemoUITests.swift
```diff
@@ -0,0 +1,50 @@
...
+        app.collectionViews.staticTexts["Pretty pictures"].tap()
+        XCTAssertTrue(app.navigationBars["Pretty Pictures"].exists)
+
+        // Tap the 2nd cell
+        let collectionView = rootCollectionView(for:app)
+        collectionView.cells.collectionViews.children(matching: .cell).element(boundBy: 0).otherElements.children(matching: .image).element.tap()
+    }
+
+    private func rootCollectionView(for app:XCUIApplication) -> XCUIElement {
+        return app.otherElements.containing(.navigationBar, identifier:"Pretty Pictures").children(matching: .other).element.children(matching: .other).element.children(matching: .other).element.children(matching: .collectionView).element
```
JohnSundell: This is a very long statement. Maybe it can be split up into multiple lines for easier readability? 🤔

demo/tests/HubFrameworkDemoUITests.swift
```diff
@@ -0,0 +1,50 @@
...
+        // Tap "Pretty pictures" and make sure we navigate to that page.
+        app.collectionViews.staticTexts["Pretty pictures"].tap()
+        XCTAssertTrue(app.navigationBars["Pretty Pictures"].exists)
+
+        // Tap the 2nd cell
+        let collectionView = rootCollectionView(for:app)
+        collectionView.cells.collectionViews.children(matching: .cell).element(boundBy: 0).otherElements.children(matching: .image).element.tap()
+    }
+
+    private func rootCollectionView(for app:XCUIApplication) -> XCUIElement {
```
JohnSundell: 🎨 Space between parameter label (`app`) and type (`XCUIApplication`).

demo/tests/HubFrameworkDemoUITests.swift
cerihughes: Agreed, but this is how Xcode generates these tests. It actually takes longer to format the generated code into readable lines than it does to generate the test code :O

I will tidy this up though...

cerihughes:
@rastersize @JohnSundell @nataliq UI Tests are now passing, but there's a problem with code coverage which *may* be the reason why failures are not reported. Looks like this is a known problem with `codecov` - see https://github.com/codecov/example-swift/issues/16

So, the question is, should we merge these tests, even though we're not sure if the build will still pass if they fail? i.e. can we merge this without https://github.com/spotify/HubFramework/issues/127 being addressed?

JohnSundell:
@cerihughes I personally think we should get it in. It's better than what we have, and it lays some of the ground work for additional testing we want to do ASAP. So while we look into why the build doesn't report as failed when the test fail, I would say we should go ahead with this 😄

JohnSundell:
@cerihughes One thing we should do before mering this though, is to add a linting rule for the license headers in the `/tests` folder of the demo app. See how I did it for the live editing code here: https://github.com/spotify/HubFramework/pull/118/commits/325fa1e365365e4db06b7213335513e70681ce2d

rastersize:
I have an idea of how to mitigate the “test failure reported as green”. Will need to test it first though.

---

Before merging this we also need to make sure Danger reports back the status of the UI tests. So we need to addition to the [`Dangerfile`](https://github.com/spotify/HubFramework/blob/master/Dangerfile).

Basically duplication of these things:
- https://github.com/spotify/HubFramework/blob/master/Dangerfile#L31
- https://github.com/spotify/HubFramework/blob/master/Dangerfile#L34-L41

JohnSundell:
@cerihughes Need any assistance on this one? Would be happy to help 😄

cerihughes:
@JohnSundell @rastersize Implemented the changes you requested.
Also, the test failures now seem to be reporting properly. I put in a breaking change, the tests failed and the build went red.

PR #126, opened 2016-11-03
Implement equality for icons.

cerihughes:
Previously, 2 equivalent `HUBIconImplementation`s would return `NO` for `isEqual:`.


cerihughes:
@Goos @krris @JohnSundell

PR #130, opened 2016-11-08
Fix components background color gets overridden when added to HubContainerView

Lumenela:
Make HubContainerView set background color only for the collection view it contains instead of overriding background color of all the subviews.

### DIFF
--- a/sources/HUBContainerView.h
+++ b/sources/HUBContainerView.h
@@ -26,7 +26,11 @@ NS_ASSUME_NONNULL_BEGIN
 /// View that acts as a container view for a Hub Framework view controller
 @interface HUBContainerView : UIView
 
-/// Collection view contained by the container
+/**
+ *  Collection view contained by the container.
+ *
+ *  @discussion When collectionView is set it's also added as a subview.
+ */
 @property (nonatomic, strong, nullable) UICollectionView *collectionView;
 
 @end
--- a/sources/HUBContainerView.h
+++ b/sources/HUBContainerView.h
@@ -26,6 +26,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// View that acts as a container view for a Hub Framework view controller
 @interface HUBContainerView : UIView
 
+/// Collection view contained by the container
+@property (nonatomic, strong, nullable) UICollectionView *collectionView;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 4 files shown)


sources/HUBContainerView.h
```diff
@@ -26,6 +26,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// View that acts as a container view for a Hub Framework view controller
 @interface HUBContainerView : UIView
 
+/// Collection view contained by the container
```
JohnSundell: Suggestion: Add a note that setting this property also adds the collection view to the container view as a subview?

sources/HUBContainerView.m
```diff
@@ -25,18 +25,28 @@
...
 @implementation HUBContainerView
 
-- (void)setBackgroundColor:(nullable UIColor *)backgroundColor
+- (void)setCollectionView:(nullable UICollectionView *)collectionView
 {
-    [super setBackgroundColor:backgroundColor];
-    
-    for (UIView * const view in self.subviews) {
-        view.backgroundColor = backgroundColor;
+    if (collectionView == nil) {
```
JohnSundell: You can avoid some code duplication here by removing the previous collection view before adding the new one (regardless of if the new one is `nil`).

sources/HUBViewControllerImplementation.m
```diff
@@ -173,8 +173,11 @@ - (void)dealloc
 
 - (void)loadView
 {
-    self.view = [[HUBContainerView alloc] initWithFrame:CGRectZero];
+    HUBContainerView *hubContainerView = [[HUBContainerView alloc] initWithFrame:CGRectZero];
+    self.view = hubContainerView;
+
     [self createCollectionViewIfNeeded];
+    hubContainerView.collectionView = self.collectionView;
```
JohnSundell: Move this into `createCollectionViewIfNeeded` instead? That way, when the collection view is re-created (for example after a memory warning), the container view's reference will be updated.

sources/HUBViewControllerImplementation.m
```diff
@@ -865,6 +866,9 @@ - (void)createCollectionViewIfNeeded
     collectionView.delegate = self;
     
     [self.view insertSubview:collectionView atIndex:0];
```
JohnSundell: You're now already doing this in `HUBContainerView`. So remove this to avoid doing this twice.

PR #131, opened 2016-11-08
Ignore warnings for Xcode 8.2

rastersize:
Since the beta is out it’d be nice to make the framework work with it 😋

Like the recently merged PR but also for Xcode 8.2.

@spotify/objc-dev


rastersize:
@JohnSundell I wish!

JensAyton:
You can merge them into `WARNING_CFLAGS` if there isn’t a use case for supporting Xcode < 7.3.0.

PR #132, opened 2016-11-08
View model loader: Support appending paginated content

JohnSundell:
This change introduces an API for appending paginated content to a view model. By making a content operation conform to `HUBContentOperationWithPaginatedContent`, new content can now be added to an existing view model builder.

To support the declarative nature of the Hub Framework, content operations are always guaranteed to execute from a “clean state”. `HUBViewModelLoader` keeps snapshots of previous view model builders & errors, to be able to make this guarantee.

However, with pagination, we don’t want to execute from a clean state. Instead, we want to append content to the **current** state, which imposes some interesting challenges on the content loading code.

To accommodate these 2 different uses cases, a new `HUBContentOperationExecutionInfo` is introduced. Instead of enqueueing operations directly, instances of this class is now used to keep track of a view model loader’s operation queue. This enables us to reference the same content operation using different modes, and to use different snapshots depending on what mode an operation is being executed in.

Went all in with the tests this time, to make sure that A) this doesn’t break any existing content loading code, and B) that all the edge cases around managing both replacing and appended state changes at the same time.

Note that the documentation for `HUBContentOperationWithPaginatedContent` is lying a bit at the moment, since the view controller hasn't been wired up to using this yet - but wanted to save that for a subsequent PR.

Provides the model layer side of https://github.com/spotify/HubFramework/issues/28

### DIFF
--- a/sources/HUBViewModelLoaderImplementation.m
+++ b/sources/HUBViewModelLoaderImplementation.m
@@ -24,6 +24,7 @@
 #import "HUBFeatureInfo.h"
 #import "HUBConnectivityStateResolver.h"
 #import "HUBContentOperationWithInitialContent.h"
+#import "HUBContentOperationWithPaginatedContent.h"
 #import "HUBContentOperationActionObserver.h"
 #import "HUBContentOperationActionPerformer.h"
 #import "HUBActionPerformer.h"
@@ -32,6 +33,7 @@
 #import "HUBViewModelBuilderImplementation.h"
 #import "HUBViewModelImplementation.h"
 #import "HUBContentOperationWrapper.h"
+#import "HUBContentOperationExecutionInfo.h"
 #import "HUBUtilities.h"
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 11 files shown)


JohnSundell:
@spotify/objc-dev

tests/HUBViewModelLoaderTests.m
```diff
@@ -674,6 +674,254 @@ - (void)testNilReloadPolicyAlwaysResultingInReload
...
+    [self.loader loadViewModel];
+    [self.loader loadNextPageForCurrentViewModel];
+    [self.loader loadNextPageForCurrentViewModel];
+    [self.loader loadNextPageForCurrentViewModel];
+    
+    id<HUBContentOperationDelegate> const contentOperationDelegate = contentOperation.delegate;
+    [contentOperationDelegate contentOperationDidFinish:contentOperation];
+    [contentOperationDelegate contentOperationDidFinish:contentOperation];
+    
+    // Make sure that we don't have any component models until the last operation is finished
```
JohnSundell: Well, we trust the API user. If they tell us to load multiple pages at once - we will. If not, they should load a new page, let Hubs render it - and then load another one. We want appending content using pagination to work in the same way as the normal content loading chain - that is - we only render once the queue is empty.

Goos: Gotcha 👍

sources/HUBViewModelLoaderImplementation.m
```diff
@@ -324,6 +368,26 @@ - (void)contentOperationQueueDidBecomeEmpty
...
+        return existingOperationWrapper;
+    }
+    
+    id<HUBContentOperation> const operation = self.contentOperations[operationIndex];
+    HUBContentOperationWrapper * const newOperationWrapper = [[HUBContentOperationWrapper alloc] initWithContentOperation:operation index:operationIndex];
+    newOperationWrapper.delegate = self;
+    self.contentOperationWrappers[@(operationIndex)] = newOperationWrapper;
+    
+    if ([operation conformsToProtocol:@protocol(HUBContentOperationWithPaginatedContent)]) {
+        self.anyContentOperationSupportsPagination = YES;
```
Goos: Awesome :smile:

PR #133, opened 2016-11-09
View controller: Recognize pan gestures on a view level

JohnSundell:
This change moves the gesture recognizer used for pan (scroll) events in the collection view to the view level. This is done in order to support scrolling on a header or overlay component, without having to resort to manual -pointInside checks.

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -258,8 +258,14 @@ NS_ASSUME_NONNULL_BEGIN
         return;
     }
 
-    [self.collectionView removeFromSuperview];
-    self.collectionView = nil;
+    if (self.collectionView != nil) {
+        UICollectionView * const collectionView = self.collectionView;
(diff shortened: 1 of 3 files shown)


JohnSundell:
@spotify/objc-dev

sources/HUBViewControllerImplementation.m
```diff
@@ -864,6 +870,7 @@ - (void)createCollectionViewIfNeeded
     collectionView.dataSource = self;
     collectionView.delegate = self;
     
+    [self.view addGestureRecognizer:collectionView.panGestureRecognizer];
```
Goos: Could this ever cause the gesture to be recognized twice? E.g. once in the UICollectionView, and once in its superview?

JohnSundell: Nope, since a gesture recognizer can only be part of a single view, it will be removed from the collection view 🙂

PR #138, opened 2016-11-10
View controller: Add API to disable scrolling

JohnSundell:
This change adds a new delegate method for `HUBViewController` that gets called every time the user is about to start scrolling a view. The delegate then has an opportunity to veto wether the scroll event should start.


JohnSundell:
@spotify/objc-dev

PR #139, opened 2016-11-10
Removes the calls to the scroll handler when scrolling manually

Goos:
Removes the calls to the scroll handler when scrolling programmatically with the `scrollToComponentOfType:...`-method. This is done to be more in line with the UICollectionView / UIScrollView interfaces, where the scrolling methods don't call the same delegate methods as scrolling by hand.


Goos:
Oops, that's right 😪

PR #140, opened 2016-11-10
Update content inset once the view model finishes rendering

Goos:
Adds another check for the content inset once the view model finishes rendering, as the recent move of header configuration introduced an inconcistency for scroll handlers where the content inset was dependant on the size of content (e.g. component frames).

@spotify/objc-dev, @cerihughes, @JohnSundell

### DIFF
--- a/tests/HUBViewControllerTests.m
+++ b/tests/HUBViewControllerTests.m
@@ -1511,6 +1511,33 @@
     XCTAssertTrue(self.collectionView.appliedScrollViewOffsetAnimatedFlag);
 }
 
+- (void)testRenderingUpdatesContentInsetBeforeAndAfterRendering
+{
+    UIEdgeInsets const firstInsets = UIEdgeInsetsMake(100, 30, 40, 200);
+    UIEdgeInsets const secondInsets = UIEdgeInsetsMake(50, 0, 0, 0);
+
+    __weak HUBViewControllerTests *weakSelf = self;
+    void (^assertInsetsEqualToCollectionViewInsets)(UIEdgeInsets insets) = ^(UIEdgeInsets insets) {
+        HUBViewControllerTests *strongSelf = weakSelf;
+        XCTAssertTrue(UIEdgeInsetsEqualToEdgeInsets(strongSelf.collectionView.contentInset, insets));
+    };
+    
+    __block NSUInteger numberOfContentInsetCalls = 0;
+    self.scrollHandler.contentInsetHandler = ^UIEdgeInsets(UIViewController<HUBViewController> *controller, UIEdgeInsets insets) {
+        numberOfContentInsetCalls += 1;
+        if (numberOfContentInsetCalls == 1) {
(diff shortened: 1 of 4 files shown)


tests/HUBViewControllerTests.m
```diff
@@ -2536,4 +2559,9 @@ - (void)registerAndGenerateComponentsWithNamespace:(NSString *)namespace
     };
 }
 
+- (void)assertEdgeInsetsEqualInsets:(UIEdgeInsets)insets
```
JohnSundell: Taken out of context this method name looks a bit off. I would suggest either just using a local block to perform these calculations, or naming it a bit more descriptive (like `assertCollectionViewContentInsetEqualToEdgeInsets`). What do you think?

tests/HUBViewControllerTests.m
```diff
@@ -1511,6 +1511,29 @@ - (void)testSetScrollOffsetForwardsAnimatedFlagToCollectionView
...
+- (void)testRenderingUpdatesContentInsetBeforeAndAfterRendering
+{
+    UIEdgeInsets const firstInsets = UIEdgeInsetsMake(100, 30, 40, 200);
+    UIEdgeInsets const secondInsets = UIEdgeInsetsMake(50, 0, 0, 0);
+    __block NSUInteger calls = 0;
+    
+    __weak HUBViewControllerTests *weakSelf = self;
+    self.scrollHandler.contentInsetHandler = ^UIEdgeInsets(UIViewController<HUBViewController> *controller, UIEdgeInsets insets) {
+        HUBViewControllerTests *strongSelf = weakSelf;
+        if (++calls == 1) {
```
JohnSundell: I can see why they removed these operators from Swift now 😅 This is a bit hard to read, could you make the increment outside of the `if` statement instead? 😇

tests/HUBViewControllerTests.m
```diff
@@ -1511,6 +1511,29 @@ - (void)testSetScrollOffsetForwardsAnimatedFlagToCollectionView
     XCTAssertTrue(self.collectionView.appliedScrollViewOffsetAnimatedFlag);
 }
 
+- (void)testRenderingUpdatesContentInsetBeforeAndAfterRendering
+{
+    UIEdgeInsets const firstInsets = UIEdgeInsetsMake(100, 30, 40, 200);
+    UIEdgeInsets const secondInsets = UIEdgeInsetsMake(50, 0, 0, 0);
+    __block NSUInteger calls = 0;
```
JohnSundell: Name this a bit more descriptively?

tests/HUBViewControllerTests.m
Goos: Yeah it's a workaround to not being able to use `XCTAssert`s while inside the content inset blocks, since that macro uses self inside 😑  but using another block makes more sense, I'll do that

JohnSundell: You can always just use `NSAssert` :)

PR #142, opened 2016-11-10
View controller: Don’t reconfigure header/overlay components serving the same model

JohnSundell:
Similar to how we deal with body components (with diffing), there’s no need to reconfigure a header or overlay component if it’s going to serve the exact same model as its currently serving.

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -1065,6 +1065,10 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState
 - (HUBComponentWrapper *)configureHeaderOrOverlayComponentWrapperWithModel:(id<HUBComponentModel>)componentModel
                                                   previousComponentWrapper:(nullable HUBComponentWrapper *)previousComponentWrapper
 {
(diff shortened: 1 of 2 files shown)


JohnSundell:
@spotify/objc-dev

cerihughes:
@JohnSundell Is this just a short-term solution / quick fix? Would a better / longer term solution be to consider headers and overlays in the existing diffing code and return diffed results for each component type?

JohnSundell:
@cerihughes Yeah it would probably be a good idea to move all code doing any diffing into `HUBViewModelDiff`. For now there's already quite an amout of "diffing code" (although very simple) in the view controller for overlay components - would love to refactor that - but let's do it in a separate PR 😄

JohnSundell:
@cerihughes With this we also have the tests in place to make sure that our future refactor won't break anything 🚀

cerihughes:
@JohnSundell - Yeah, that makes sense. Let's create an issue to move all diffing code into the same place.

JohnSundell:
@cerihughes Done! https://github.com/spotify/HubFramework/issues/144

PR #145, opened 2016-11-10
Extend HubViewModel builder with method to set value for key in customData

Lumenela:
The method makes mutable copy of customData, sets value for the key or deletes value for the key if the value is nil and finally sets updated dictionary to self.customData

### DIFF
--- a/sources/HUBViewModelBuilderImplementation.m
+++ b/sources/HUBViewModelBuilderImplementation.m
@@ -266,6 +266,22 @@ NS_ASSUME_NONNULL_BEGIN
                                                        customData:[self.customData copy]];
 }
 
+#pragma mark - Manipulate custom data
+
+- (void)setCustomDataValue:(nullable id)value forKey:(nonnull NSString *)key
+{
+    NSMutableDictionary *customData = self.customData.mutableCopy;
+    customData = customData ?: [[NSMutableDictionary alloc] init];
+
+    if (value == nil) {
+        [customData removeObjectForKey:key];
+    } else {
+        [customData setObject:(id)value forKey:key];
+    }
+
+    self.customData = customData;
+}
+
 #pragma mark - HUBJSONCompatibleBuilder
 
 - (nullable NSError *)addJSONData:(NSData *)JSONData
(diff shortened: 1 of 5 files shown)


sources/HUBViewModelBuilderImplementation.m
```diff
@@ -266,6 +266,22 @@ - (void)removeAllComponentModelBuilders
                                                        customData:[self.customData copy]];
 }
 
+#pragma mark - Manipulate custom data
+
+- (void)setCustomDataValue:(nullable id)value forKey:(nonnull NSString *)key
+{
+    NSMutableDictionary *customData = self.customData.mutableCopy;
```
JohnSundell: Since mutable copy is an action being taken on the dictionary, call it as a method with `[]` notation.

sources/HUBViewModelBuilderImplementation.m
```diff
@@ -266,6 +266,22 @@ - (void)removeAllComponentModelBuilders
...
+    customData = customData ?: [[NSMutableDictionary alloc] init];
```
JohnSundell: Suggestion: You could avoid `customData` becoming `nullable` by combining this statement with the one above:
```objective-c
NSMutableDictionary * const customData = [self.customData mutableCopy] ?: [NSMutableDictionary new];
```

JohnSundell: Or maybe the compiler won't let you do that? 🤔

tests/HUBViewModelBuilderTests.m
```diff
@@ -73,6 +73,17 @@ - (void)testPropertyAssignment
     XCTAssertEqualObjects(model.customData, self.builder.customData);
 }
 
+- (void)testCustomDataUpdates
+{
+    [self.builder setCustomDataValue:@"test_object" forKey:@"test_key"];
+
+    XCTAssertEqual([self.builder.customData objectForKey:@"test_key"], @"test_object");
```
JohnSundell: Use `XCTAssertEqualObjects` to avoid comparing against the exact same string instance.

PR #146, opened 2016-11-10
Serialise the group attribute

cerihughes:
... also adds tests that test the full serialisation / deserialisation stack for the whole schema.


cerihughes:
@JohnSundell

PR #147, opened 2016-11-10
Fix potential crashers in unit tests

rastersize:
`XCTestExpectation` is a bit dangerous since it might crash the test if it’s fulfilled after `waitForExpectationWithTimeout:handler:` times out. The workaround is to make the variable referencing the expectation weak. So that it goes away and the `-fulfill` message is instead sent to `nil`.

@spotify/objc-dev

### DIFF
--- a/tests/HUBComponentWrapperTests.m
+++ b/tests/HUBComponentWrapperTests.m
@@ -94,7 +94,7 @@
     XCTAssertEqual(self.selectionStateFromWillUpdateDelegateMethod, HUBComponentSelectionStateHighlighted);
     XCTAssertEqual(self.selectionStateFromDidUpdateDelegateMethod, HUBComponentSelectionStateNone);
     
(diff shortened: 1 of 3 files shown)


JohnSundell:
@rastersize Wouldn't we want it to crash if something went wrong and the expectation wasn't fulfilled? 🤔

rastersize:
@JohnSundell No we don’t want it to crash since it could, and should, just fail the test instead. Allowing the rest of the suite to run. Which will happen after this PR. The current behavior will be that the tests, if there’s for example CPU congestion, crashes and no other tests will be ran.

What happens is a edge case of `XCTestExpectation`’s API. For reference see http://stackoverflow.com/questions/27555499/xctestexpectation-how-to-avoid-calling-the-fulfill-method-after-the-wait-contex/27555500#27555500

8W9aG:
👍

Issue #149, opened 2016-11-11
Update Danger to check that the test results exists.


cerihughes:
We sometimes see this error in the build:
```
[17:31:02]: ▸ 2016-11-10 17:31:02.309 xcodebuild[6305:14220] Error Domain=FBSOpenApplicationServiceErrorDomain Code=1 "The request to open "com.apple.test.HubFrameworkDemoUITests-Runner" failed." UserInfo={NSUnderlyingError=0x7fc0f3c22d90 {Error Domain=FBSOpenApplicationErrorDomain Code=4 "Application "com.apple.test.HubFrameworkDemoUITests-Runner" is unknown to FrontBoard." UserInfo={BSErrorCodeDescription=NotFound, NSLocalizedFailureReason=Application "com.apple.test.HubFrameworkDemoUITests-Runner" is unknown to FrontBoard.}}, NSLocalizedDescription=The request to open "com.apple.test.HubFrameworkDemoUITests-Runner" failed., NSLocalizedFailureReason=The request was denied by service delegate (SBMainWorkspace) for reason: NotFound ("Application "com.apple.test.HubFrameworkDemoUITests-Runner" is unknown to FrontBoard")., BSErrorCodeDescription=RequestDenied}
```

We should update the `Dangerfile` to check for `TestSummaries.junit` before trying to parse it.

e.g. something like
```
def report_junit_results(path)
  junit_report_path = Dir.glob(path + "/*TestSummaries.junit").first
  if File.readable_real?(junit_report_path)
    junit.parse junit_report_path
    junit.report
  else
    error "Couldn't find the (junit) unit test report file in " + path + ". Make sure the tests were actually run."
  end
end
```

cerihughes:
Review at https://github.com/spotify/HubFramework/pull/150

PR #150, opened 2016-11-11
Check the junit file is readable

cerihughes:
Make sure the junit result file is readable before attempting to parse it.

Addresses https://github.com/spotify/HubFramework/issues/149


cerihughes:
@rastersize : I copied your suggested changes into this PR. Look good?

PR #155, opened 2016-11-15
Fix layout crash when a hub feature is presented while backgrounded

ShervinS:
Solves that when a hub view controller is presented while an app is in the background, the collectionview layout asserts on app foregrounding.

@Goos @JohnSundell

### DIFF
--- a/sources/HUBViewControllerImplementation.m
+++ b/sources/HUBViewControllerImplementation.m
@@ -392,7 +392,7 @@ NS_ASSUME_NONNULL_BEGIN
     }
     
     NSIndexPath * const indexPath = [NSIndexPath indexPathForItem:(NSInteger)index inSection:0];
-    return [self.collectionView layoutAttributesForItemAtIndexPath:indexPath].frame;
+    return [self.collectionView.collectionViewLayout layoutAttributesForItemAtIndexPath:indexPath].frame;
 }
 
 - (NSUInteger)indexOfBodyComponentAtPoint:(CGPoint)point


sources/HUBViewControllerImplementation.m
```diff
@@ -392,7 +392,8 @@ - (CGRect)frameForBodyComponentAtIndex:(NSUInteger)index
     }
     
     NSIndexPath * const indexPath = [NSIndexPath indexPathForItem:(NSInteger)index inSection:0];
-    return [self.collectionView layoutAttributesForItemAtIndexPath:indexPath].frame;
+    HUBCollectionViewLayout * const layout = (HUBCollectionViewLayout *)self.collectionView.collectionViewLayout;
```
JohnSundell: Why do you need this casting?

ShervinS: I could skip the casting, I just added it for clarity to understand that it differs from a UICollectionViewLayout. 

If you call `layoutAttributesForItemAtIndexPath` on a collectionView directly, the collection view asserts if there is no layout for that indexpath.

But in our own `HUBCollectionViewLayout` class, we do this: 
```objc
- (nullable UICollectionViewLayoutAttributes *)layoutAttributesForItemAtIndexPath:(NSIndexPath *)indexPath
{
    return self.layoutAttributesByIndexPath[indexPath];
}
```

which just returns nil. 
And this method can be called before the layout is done if called while the app is in the background. By using our method, we just return nil instead of asserting.

ShervinS:
I'd really like a test for this, but it's hard with the app state going to the background / foreground. 
We're adding TA tests for it on our side, but if you can think of a nice way to unit test it I'd be glad to :)

Goos:
@JohnSundell the cause of the crash is related to the fix I did last week (https://github.com/spotify/HubFramework/pull/141) where the bounds-checking was changed as the view model child count might not align with the collection view's item count at all times. 

However, it seems like the `numberOfItemsInSection:` will actually trigger a call to the data source if the collection view hasn't done any prior item counts, potentially causing the collection view to have more items than there are layout attributes for (as the UICollectionViewLayout is computed manually in combination with the reloading, which hasn't occurred yet). 

So the crash would remain if `frameForBodyComponentAtIndex:` was called after the view model was updated but _before the first rendering pass_. 

Yay UIKit \o/

sources/HUBViewControllerImplementation.m
JohnSundell: Ah, I understand. You should be able to get rid of the casting sine you're just calling a `UICollectionViewLayout` API.

ShervinS:
Removed the typecast @JohnSundell :)

PR #156, opened 2016-11-15
Changes how the completion handler in the scrolling method is called.

Goos:
Changes the callback in the `scrollToComponentOfType:indexPath:scrollPosition:animated:completion:` so that it is called when each step of the scrolling is completed, rather than once after all the steps.

@spotify/objc-dev


Goos:
[15:09:30]: ▸ ❌  /Users/travis/build/spotify/HubFramework/sources/HUBViewControllerImplementation.m:1476:38: variable length array used [-Werror,-Wvla]

Don't we support variable-length arrays? It's been in the standard for a while 😛

dflems:
@goos: Yes and no. VLA's *can* be dangerous. They (may) allocate space on the stack, so you need to be careful. For small value, it's very efficient, but if you inadvertently pass a large value in, it can have consequences.

In this particular case it's likely fine, but you could probably also get away with a fixed-length buffer and an assertion. Index path lengths usually aren't particularly long.

Goos:
@dflems good point – but doesn't that apply for most things C? E.g. sending an unintentionally large size to malloc :smile:

dflems:
@goos: True, true. In this case it definitely seems fine.

Also, please remove the project changes and add this to `WARNING_CFLAGS` in `spotify_os.xcconfig` (and push [upstream](https://github.com/spotify/ios-ci)).

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -1334,6 +1334,19 @@
 				GCC_DYNAMIC_NO_PIC = NO;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				SDKROOT = iphoneos;
+				WARNING_CFLAGS = (
```
rastersize: All warnings should be added via the [`spotify_os.xcconfig`](https://github.com/spotify/HubFramework/blob/master/spotify_os.xcconfig) otherwise we can’t override the settings internally.

Issue #158, opened 2016-11-16
Expose `HUBViewController` as a concrete class


JohnSundell:
Currently, the view controller of the Hub Framework is exposed in the API as a protocol (`HUBViewController`), and when a view controller is created it's returned as type `UIViewController<HUBViewController>`. This works great in Objective-C but is not compatible with Swift, since a type can't be both a protocol and a concrete type at the same time in Swift.

To fix this, we need to expose `HUBViewController` as a class, rather than a protocol. That way we can have `HUBViewControllerFactory` return objects of type `HUBViewController *`, making it compatible with both Objective-C and Swift.

This change will also enable API users to create view controllers directly, if they so choose.

This requires a bit of refactoring, as all initializer parameters of `HUBViewController` need to have types that are available in the public API.

JohnSundell:
Implemented and shipped in `2.0.6`

PR #159, opened 2016-11-16
Use HUBActionHandlerWrapper as HUBActionHandler protocol

JohnSundell:
This change is part of the work for https://github.com/spotify/HubFramework/issues/158, which has the end goal of making `HUBViewController` publicly initializable. To do this, we need to remove all dependencies on concrete internal classes.

This is the first step, which makes `HUBViewController` take any `HUBActionHandler` as part of its initializer, rather than relying on the concrete `HUBActionHandlerWrapper` implementation. To make this happen, logic that was previously delegated to the view controller around chaining of async actions needed to be moved into the action handler wrapper itself.

The test change that was made in `HUBViewControllerTests` needed to make made to avoid implicit autorelease pools from taking over the lifecycle of objects.

### DIFF
--- a/sources/HUBActionHandlerWrapper.m
+++ b/sources/HUBActionHandlerWrapper.m
@@ -121,7 +121,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)actionDidFinish:(HUBAsyncActionWrapper *)action
         withContext:(id<HUBActionContext>)context
         chainToActionWithIdentifier:(nullable HUBIdentifier *)nextActionIdentifier
-        customData:(nullable NSDictionary<NSString *,id> *)nextActionCustomData
+        customData:(nullable NSDictionary<NSString *, id> *)nextActionCustomData
 {
     [self.ongoingAsyncActions removeObject:action];
     
--- a/sources/HUBActionHandlerWrapper.m
+++ b/sources/HUBActionHandlerWrapper.m
@@ -22,9 +22,10 @@
 #import "HUBActionHandlerWrapper.h"
 
 #import "HUBActionRegistryImplementation.h"
-#import "HUBActionContext.h"
+#import "HUBActionContextImplementation.h"
 #import "HUBAction.h"
 #import "HUBAsyncAction.h"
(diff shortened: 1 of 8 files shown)


JohnSundell:
@spotify/objc-dev

sources/HUBActionHandlerWrapper.m
```diff
@@ -114,23 +116,26 @@ - (BOOL)handleActionWithContext:(id<HUBActionContext>)context
...
 
-#pragma mark - HUBAsyncActionDelegate
+#pragma mark - HUBAsyncActionWrapperDelegate
 
-- (void)actionDidFinish:(id<HUBAsyncAction>)action
+- (void)actionDidFinish:(HUBAsyncActionWrapper *)action
+        withContext:(id<HUBActionContext>)context
         chainToActionWithIdentifier:(nullable HUBIdentifier *)nextActionIdentifier
-        customData:(nullable NSDictionary<NSString *, id> *)nextActionCustomData
+        customData:(nullable NSDictionary<NSString *,id> *)nextActionCustomData
```
krris: Shouldn't be a space between `NSString` and `id`? 🔬

sources/HUBAsyncActionWrapper.h
```diff
@@ -0,0 +1,53 @@
...
+ *  Notify an async action wrapper's delegate that it finished
+ *
+ *  @param action The action that finished
+ *  @param context The context that the action was performed using
+ *  @param nextActionIdentifier The identifier of any chained action to perform
+ *  @param nextActionCustomData Any custom data to pass to the chained action
+ */
+- (void)actionDidFinish:(HUBAsyncActionWrapper *)action
+            withContext:(id<HUBActionContext>)context
+        chainToActionWithIdentifier:(nullable HUBIdentifier *)nextActionIdentifier
```
krris: Strange indentation but there's no way to fix this 😅

sources/HUBActionHandlerWrapper.m
JohnSundell: There should! Great catch 👍

PR #160, opened 2016-11-16
View controller: Use component registry as HUBComponentRegistry protocol

JohnSundell:
This change is part of the work for https://github.com/spotify/HubFramework/issues/158, which has the end goal of making `HUBViewController` publicly initializable. To do this, we need to remove all dependencies on concrete internal classes.

With this, the API to create new component instances has been moved from `HUBComponentRegistryImplementation` to the `HUBComponentRegistry` API. While it’s not expected that any API user will use this method directly, it enables the view controller to not rely on the concrete component registry implementation.


JohnSundell:
@spotify/objc-dev

PR #167, opened 2016-11-17
Expose HUBViewController as a concrete class

JohnSundell:
This change makes `HUBViewController` a concrete class (instead of an API protocol/implementation class pair) and exposes it as part of the public API. The initializer is still kept private (for now), since it relies on some dependencies being passed down from `HUBViewControllerFactory`.

This change is made to be more compatible with Swift, where currently `HUBViewControllerFactory` returns `UIViewController` as a type (for more info - see https://github.com/spotify/HubFramework/issues/158).


JohnSundell:
@spotify/objc-dev

JohnSundell:
Mr @Goos you are one thorough reviewer! 👍 So awesome 🚀 Will fix!

Goos:
@JohnSundell `git grep` is the shizzle :smile:

PR #169, opened 2016-11-18
Demo app: Add a “sticky header” component + feature and use it for UI tests

JohnSundell:
This change adds a header component that applies a parallax-like effect to its background image. It’s showcased in a new “Sticky header” feature.

This is done partly for demonstration purposes, but mostly to be able to perform UI tests against a header component, to assert that the framework behaves the correct way when one is used.

To avoid further code duplication in UI tests, a `UITestCase` base class is created that performs the “boilerplate” operations involved in setting up a UI test.

This PR now also includes a bug fix for content insets, where the content insets of a view could become invalid if a view containing a header component was reloaded.


JohnSundell:
@spotify/objc-dev

demo/sources/NavigationController.swift
```diff
@@ -0,0 +1,61 @@
...
+    func viewControllerShouldStartScrolling(_ viewController: HUBViewController) -> Bool {
+        return true
+    }
+    
+    func viewController(_ viewController: HUBViewController, componentWith componentModel: HUBComponentModel, layoutTraits: Set<HUBComponentLayoutTrait>, willAppearIn componentView: UIView) {
+        // No-op
+    }
+    
+    func viewController(_ viewController: HUBViewController, componentWith componentModel: HUBComponentModel, layoutTraits: Set<HUBComponentLayoutTrait>, didDisappearFrom componentView: UIView) {
+
```
krris: `// No-op` is missing?

JohnSundell:
@krris Added another test, and fixed a bug affecting the content inset calcuation when a header component is used. Wanna take a 2nd look? 😄

PR #170, opened 2016-11-20
Improvement: stick to the convention for performing on main queue.

krris:
Small cleaning: instead of calling `dispatch_async(dispatch_get_main_queue(), ^{})` we should use `HUBPerformOnMainQueue` just like in other places of the framework.


sources/HUBViewController.m
```diff
@@ -1258,7 +1258,7 @@ - (void)loadImageFromData:(id<HUBComponentImageData>)imageData
         contextsForURL = [NSMutableArray arrayWithObject:context];
         self.componentImageLoadingContexts[imageURL] = contextsForURL;
         
-        dispatch_async(dispatch_get_main_queue(), ^{
+        HUBPerformOnMainQueue(^{
```
JohnSundell: Actually, here we want to always run the code async, to make sure that the cell that the image is for is registered with `UICollectionView` before it is loaded. We should add a test for that, that would've failed this change.

sources/HUBViewController.m
```diff
@@ -1490,7 +1490,7 @@ - (void)scrollToRemainingComponentsOfType:(HUBComponentType)componentType
             /* This solves a case where the UICollectionView hasn't updated its visible cells until the next cycle
                when changing the content offset without animations. */
             if (!animated) {
-                dispatch_async(dispatch_get_main_queue(), stepCompletionHandler);
+                HUBPerformOnMainQueue(stepCompletionHandler);
```
JohnSundell: Same thing here. We want to do it async (the comment here says why)

JohnSundell:
@krris Could you make this PR only contain the "valid" fix, so we can merge it in? 😄

krris:
@JohnSundell

PR #173, opened 2016-11-21
Make it possible for body component to start behind header component.

Lumenela:
Extend HUBViewController delegate with a method that asks if we should ignore content insets for header component.


include/HubFramework/HUBViewController.h
```diff
@@ -135,6 +135,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)viewController:(HUBViewController *)viewController componentSelectedWithModel:(id<HUBComponentModel>)componentModel;
 
+/**
+ *  Sent to a Hub Framework view controller's delegate to ask if content inset for header
+ *  component should be ignored
+ *
+ *  @param viewController The view controller which displays header component
```
JohnSundell: Should be `displays a header component`

PR #174, opened 2016-11-21
View controller: Fix a bug that would cause an invalid header state

JohnSundell:
This includes the bug fix part of https://github.com/spotify/HubFramework/pull/169, which cannot be merged yet due to infrastructure problems 😢. Also includes a new unit test that verifies the fix.

The problem:

*The content insets of a view controller were calculated using the **current** size of a header component, rather than using the **original** size. So when a header would be collapsed, the view could end up having an invalid content inset state.*


JohnSundell:
@spotify/objc-dev

JohnSundell:
@lumenela had to move your code a bit, also made the test a bit more thorough, mind verifying that everything looks ✅? 😄

Lumenela:
@JohnSundell Looks good. I should have done it myself :)

PR #172, opened 2016-11-21
Run UI tests on Xcode 8.2

JohnSundell:
In an attempt to improve UI test stability (and avoid the dreaded `exit code 65` from `xcodebuild`), let’s try running our UI tests on Xcode 8.2, which has been suggested to include a lot of stability fixes for such tests.

PR #176, opened 2016-11-21
Use Xcode 8.1 for CI

JohnSundell:
Compatibility job using Xcode 7 still remains
Alternative to https://github.com/spotify/HubFramework/pull/172


JohnSundell:
YAY! Built on first run, gonna take that as a sign 😄

PR #178, opened 2016-11-22
Bug: contentOffsetObserver is called if the content offset hasn’t been updated.

krris:
🐞  This PR fixes following bug:
* `HUBContentOffsetObserver` is called if the content offset hasn’t been updated.

**How to reproduce**
1. Make the search bar in `GitHubFeature` in demo app a header component.
2. Hide the navigation bar.
3. Type sth in search bar.

**Current behaviour**
* Keyboard is hidden every time sth types one character.

**Expected behaviour**
* Keyboard is not hidden every time user types a letter.

**Final result after the fix**
screen shot 2016-11-22 at 3 07 47 pm

### DIFF
--- a/sources/HUBViewController.m
+++ b/sources/HUBViewController.m
@@ -412,7 +412,7 @@ NS_ASSUME_NONNULL_BEGIN
     const CGFloat x = contentOffset.x;
     const CGFloat y = contentOffset.y - self.collectionView.contentInset.top;
     
-    [self setContentOffset:CGPointMake(x, y) inCollectionView:self.collectionView animated:animated];
+    [self setContentOffset:CGPointMake(x, y) animated:animated];
 }
 
 - (void)scrollToComponentOfType:(HUBComponentType)componentType
@@ -1107,7 +1107,7 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState
         self.collectionView.contentInset = contentInsets;
         CGPoint contentOffset = self.collectionView.contentOffset;
         contentOffset.y = -contentInsets.top;
-        [self setContentOffset:contentOffset inCollectionView:self.collectionView animated:NO];
+        [self setContentOffset:contentOffset animated:NO];
     }
 
     self.collectionView.scrollIndicatorInsets = self.collectionView.contentInset;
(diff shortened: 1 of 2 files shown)


sources/HUBViewController.m
```diff
@@ -411,7 +412,7 @@ - (void)scrollToContentOffset:(CGPoint)contentOffset animated:(BOOL)animated
     const CGFloat x = contentOffset.x;
     const CGFloat y = contentOffset.y - self.collectionView.contentInset.top;
     
-    [self.collectionView setContentOffset:CGPointMake(x, y) animated:animated];
+    [self setContentOffset:CGPointMake(x, y) inCollectionView:self.collectionView animated:animated];
```
JohnSundell: Why do you need to send the collection view here? The view controller already knows about it.

sources/HUBViewController.m
```diff
@@ -1174,8 +1175,10 @@ - (void)headerAndOverlayComponentViewsWillAppear
 - (void)componentWrapperWillAppear:(HUBComponentWrapper *)componentWrapper
 {
     [componentWrapper viewWillAppear];
-    
-    if (componentWrapper.isContentOffsetObserver) {
+
+    BOOL wasContentOffsetUpdated = !self.lastContentOffset || !CGPointEqualToPoint([self.lastContentOffset CGPointValue], self.collectionView.contentOffset);
```
JohnSundell: 🎨 Compare explicitly to `nil`

sources/HUBViewController.m
krris: I was thinking about it. I wanted to make it explicit, but maybe it's better to keep functions short 😄

sources/HUBViewController.m
JohnSundell: Sure. Basically otherwise you're relying on implicit integer conversion between a pointer and `nil`, which means that you are technically comparing across different types (it only so happens to work because `nil` gets implicitly converted to `0`, both as a pointer and as a literal). This is why Swift prohibits these type of comparisons, as it has a much more static and thorough type system.

krris:
@JohnSundell 🎾

PR #179, opened 2016-11-22
Ability to ignore top bar content inset

Lumenela:
Added method to `HUBViewControllerDelegate` to ask if need to ignore top bar (nav bar + status bar) content inset.

### DIFF
--- a/include/HubFramework/HUBViewController.h
+++ b/include/HubFramework/HUBViewController.h
@@ -145,20 +145,15 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)viewController:(HUBViewController *)viewController componentSelectedWithModel:(id<HUBComponentModel>)componentModel;
 
 /**
- *  Sent to a Hub Framework view controller's delegate to ask if content inset for header
- *  component should be ignored
+ *  Sent to a Hub Framework view controller's delegate to ask if view controller should automatically
+ *  manage content inset.
  *
- *  @param viewController The view controller which displays a header component
- */
-- (BOOL)viewControllerShouldIgnoreHeaderComponentContentInset:(HUBViewController *)viewController;
-
-/**
- *  Sent to a Hub Framework view controller's delegate to ask if content inset for top bar
- *  (navigation bar and status bar) should be ignored
+ *  @param viewController The view controller which displays some components
  *
- *  @param viewController The view controller which displays a header component
+ *  @discussion When view controller automatically manages content inset it puts body components
+ *  below header component and below navigation bar.
  */
-- (BOOL)viewControllerShouldIgnoreTopBarContentInset:(HUBViewController *)viewController;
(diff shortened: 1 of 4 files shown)


include/HubFramework/HUBViewController.h
```diff
@@ -152,6 +152,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (BOOL)viewControllerShouldIgnoreHeaderComponentContentInset:(HUBViewController *)viewController;
 
+/**
+ *  Sent to a Hub Framework view controller's delegate to ask if content inset for top bar
+ *  (navigation bar and status bar) should be ignored
+ *
+ *  @param viewController The view controller which displays a header component
+ */
+- (BOOL)viewControllerShouldIgnoreTopBarContentInset:(HUBViewController *)viewController;
```
JohnSundell: Could this be merged into the API that was previously introduced for header components, and be something like `viewControllerShouldAutomaticallyManageTopContentInset:`?

Lumenela:
@JohnSundell I've replaced to delegate methods for managing content insets with one. Ready for review.

tests/HUBViewControllerTests.m
```diff
@@ -91,7 +91,7 @@ @interface HUBViewControllerTests : XCTestCase <HUBViewControllerDelegate>
 @property (nonatomic, assign) BOOL didReceiveViewControllerDidFinishRendering;
 @property (nonatomic, copy) void (^viewControllerDidFinishRenderingBlock)(void);
 @property (nonatomic, copy) BOOL (^viewControllerShouldStartScrollingBlock)(void);
-@property (nonatomic, copy) BOOL (^viewControllerShouldIgnoreHeaderComponentInset)(void);
+@property (nonatomic, copy) BOOL (^viewControllerShouldAutomaticallyManageTopContentInset)(void);
```
Goos: Just out of curiosity, how come these are blocks rather than just `BOOL` properties?

sources/HUBViewController.m
```diff
@@ -973,21 +973,26 @@ - (nullable HUBComponentWrapper *)componentWrapperFromCell:(HUBComponentCollecti
 
 - (CGFloat)calculateTopContentInset
 {
+    id<HUBViewControllerDelegate> delegate = self.delegate;
+
+    if (![delegate viewControllerShouldAutomaticallyManageTopContentInset:self]) {
```
JohnSundell: This means that if the delegate is `nil`, the view controller won't manage content inset, which is not the behavior we want (the default should be to automatically manage). So you'll need to `nil`-check the delegate first.

tests/HUBViewControllerTests.m
JohnSundell: I think we started doing it because we needed flexibility of controlling the implementation in some places, and then as multiple devs have added new tests they've stuck to that convention. I'm fine with changing these to simple `BOOL`s, but it should be done in a separate PR that doesn't contain any logic changes, for satefy 😄

PR #182, opened 2016-11-23
Dont overflow during component adjustment

aliak00:
In HUBCollectionViewLayout computeForCollectionViewSize when the only
component requires adjustment we call updateLayoutAttributesForComponentsIfNeeded
on an unsigned int with lastComponent - 1, which is, big. This then calls
updateLayoutAttributesForComponents which goes in to a loop from 0 to a bajillion.

### DIFF
--- a/sources/HUBCollectionViewLayout.m
+++ b/sources/HUBCollectionViewLayout.m
@@ -107,7 +107,7 @@ NS_ASSUME_NONNULL_BEGIN
         
         if (couldFitOnTheRow == NO) {
             [self updateLayoutAttributesForComponentsIfNeeded:componentsOnCurrentRow
-                                           lastComponentIndex:componentIndex - 1
+                                           lastComponentIndex:(NSInteger)componentIndex - 1
                                               firstComponentX:firstComponentOnCurrentRowOrigin.x
                                                lastComponentX:currentPoint.x
                                                      rowWidth:collectionViewSize.width];
@@ -155,7 +155,7 @@ NS_ASSUME_NONNULL_BEGIN
         if (isLastComponent) {
             // We center components if needed when we go to a new row. If it is the last row we need to center it here
             [self updateLayoutAttributesForComponentsIfNeeded:componentsOnCurrentRow
-                                           lastComponentIndex:componentIndex
+                                           lastComponentIndex:(NSInteger)componentIndex
                                               firstComponentX:firstComponentOnCurrentRowOrigin.x
                                                lastComponentX:currentPoint.x
                                                      rowWidth:collectionViewSize.width];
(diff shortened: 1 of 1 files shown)


aliak00:
I'm not super familiar with hub ideologies and philosophies yet so maybe I'm just doing something completely retarded, but the overflow is reproducible with this `HUBComponentLayoutManager` implementation (as long as `marginBetweenComponent(withLayoutTraits:andContentEdge)` and `horizontalOffsetForComponents(withLayoutTraits:firstComponentLeadingHorizontalOffset:lastComponentTrailingHorizontalOffset)` return the same value > 0

```
class ComponentLayoutManager: NSObject, HUBComponentLayoutManager {

    func marginBetweenComponent(
        withLayoutTraits layoutTraits: Set<HUBComponentLayoutTrait>,
        andContentEdge contentEdge: HUBComponentLayoutContentEdge
    ) -> CGFloat {
        return 1
    }

    func verticalMarginBetweenComponent(
        withLayoutTraits layoutTraits: Set<HUBComponentLayoutTrait>,
        andHeaderComponentWithLayoutTraits headerLayoutTraits: Set<HUBComponentLayoutTrait>
    ) -> CGFloat {
        return 0
    }

    func verticalMarginForComponent(
        withLayoutTraits layoutTraits: Set<HUBComponentLayoutTrait>,
        precedingComponentLayoutTraits: Set<HUBComponentLayoutTrait>
    ) -> CGFloat {
        return 0
    }

    func horizontalMarginForComponent(
        withLayoutTraits layoutTraits: Set<HUBComponentLayoutTrait>,
        precedingComponentLayoutTraits: Set<HUBComponentLayoutTrait>
    ) -> CGFloat {
        return 0
    }

    func horizontalOffsetForComponents(
        withLayoutTraits componentsTraits: [Set<HUBComponentLayoutTrait>],
        firstComponentLeadingHorizontalOffset firstComponentLeadingOffsetX: CGFloat,
        lastComponentTrailingHorizontalOffset lastComponentTrailingOffsetX: CGFloat
    ) -> CGFloat {
        return 1
    }
}
```

aliak00:
Whoops! Forgot to fastlane test_ui :) Though something tells me I'm not familiar enough with codebase for this change to the correct way to fix this.

Goos:
Hey @aliak00 – great catch, that's a pretty nasty bug! I reckon it's pretty hard to add a unit test for this since it didn't actually break before, right? 🤔 

Regarding the failing tests – you can run the UI tests by running the test target in the demo project.

aliak00:
@Goos Right. Just a lot more processing (mem use was going over 4 gigs). And yeah I think adding a unit test would be quite hard. Unsigned int underflow is "well defined" after all. I tried [clang static analysis](http://clang-analyzer.llvm.org/) on the hub and no complaints either.

And thanks! Adding the UITest to the Test section of the HubFrameworkDemo configuration made fastlane test_ui work 👍 

You'd think UI tests would not break because of this fix though :/

JohnSundell:
Hopefully UI tests flakiness will be reduced by https://github.com/spotify/HubFramework/pull/187

PR #183, opened 2016-11-23
Fix issue with scrolling method callback

Goos:
Fixes an issue with the `scrollToComponentOfType:...` method, where the callback would sometimes be called before the scrolling occured in cases where the component was already visible.

@spotify/objc-dev

### DIFF
--- a/sources/HUBViewController.m
+++ b/sources/HUBViewController.m
@@ -1407,31 +1407,25 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState
 {
     NSParameterAssert(componentIndex <= (NSUInteger)[self.collectionView numberOfItemsInSection:0]);
 
-    NSIndexPath * const rootIndexPath = [NSIndexPath indexPathForItem:(NSInteger)componentIndex inSection:0];
     CGPoint const contentOffset = [self.scrollHandler contentOffsetForDisplayingComponentAtIndex:componentIndex
                                                                                   scrollPosition:scrollPosition
                                                                                     contentInset:self.collectionView.contentInset
                                                                                      contentSize:self.collectionView.contentSize
                                                                                   viewController:self];
-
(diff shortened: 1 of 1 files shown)


sources/HUBViewController.m
```diff
@@ -1407,31 +1407,25 @@ - (void)scrollToRootBodyComponentAtIndex:(NSUInteger)componentIndex
...
-        completionWrapper();
+    // If the target offset is the same, the completion handler can be called instantly.
+    if (CGPointEqualToPoint(contentOffset, self.collectionView.contentOffset)) {
+        completion();
     // If the scrolling is animated, the animation has to end before the new component can be retrieved.
     } else if (animated) {
-        self.pendingScrollAnimationCallback = completionWrapper;
-        [self setContentOffset:contentOffset animated:animated];
+        self.pendingScrollAnimationCallback = completion;
+        [self.collectionView setContentOffset:contentOffset animated:animated];
```
JohnSundell: Think this might break the bug fix that @krris was working on yesterday, that caches content offset. Hopefully the tests will catch it 😁

krris: Josu is right. It might break this bug fix: https://github.com/spotify/HubFramework/pull/178
Please use `[self setContentOffset:contentOffset animated:animated];` instead.

JohnSundell: If the tests do not fail because of this, we should improve the tests. What do you think @krris? 😄

Goos: @JohnSundell, @krris good catch, will fix 👍

Goos:
@krris fixed 😄

sources/HUBViewController.m
krris: @JohnSundell We should definitely improve tests. It's so sad it didn't catch that 😢

PR #184, opened 2016-11-23
Add missing nullability macros HUBComponentFallbackHandler.h

rastersize:
Makes sure Swift implementors don’t get the force-unwrap of the types by default.

@spotify/objc-dev


JohnSundell:
Awesome @rastersize! 🚀

PR #186, opened 2016-11-24
Remove a component wrapper’s gesture recognizer when it’s deallocated

JohnSundell:
This patch fixes a bug that would leave gesture recognizers orphaned in their component wrapper’s superview, because they were never removed when the wrapper was deallocated.


JohnSundell:
@spotify/objc-dev

JohnSundell:
@Goos You and me both! 🎉

Issue #181, opened 2016-11-23
Make all components' (header, body + overlay) layout managed by collection view layout


JohnSundell:
Currently, only body components are managed by a `UICollectionView` and our custom `HUBCollectionViewLayout` layout object. Header- & overlay components are laid out separately, using only the component's `preferredViewSize` as input.

We should be able to use the collection view layout to drive the layout of all components, using `z indexes` to layer the components in 3 layers (from bottom: `body, header, overlay`). This would also make it possible to use `layoutTraits` for header & overlay components.

rastersize:
I think it might be nice to lay the (page) header as a special supplementary view with of the an element kind like `HUBPageHeaderKind`.

That way the collection view layout could more easily incorporate special logic for that component.

JohnSundell:
@rastersize Sure, I'm all for that 👍

JensAyton:
The Hub Framework is being phased out at Spotify, and therefore we will not be maintaining it further.

PR #187, opened 2016-11-24
Add margin instead of content inset for header component

JohnSundell:
Currently, when a header component is being used, the default behavior is to add top `contentInset` to account for its height. This brings some problems in non-Hub powered code, that can make assumptions about content inset that are no longer true. A more correct way of handling this (thanks @mhallendal for bringing this up!), would be to add proper top margin when we compute our `UICollectionViewLayout` - so that's what we now do!

A lot of tests were also making assumptions about `contentInset` that no longer are true, so those tests needed to be refactored to use component frames, insted of insets.

This is a step towards https://github.com/spotify/HubFramework/issues/181


JohnSundell:
@spotify/objc-dev

PR #189, opened 2016-11-24
Add a UI test that verifies reload functionality

JohnSundell:
Add a UI test that verifies that a view can be rapidly reloaded without any problems.

To enable this test, the "Todo list" feature in the demo app now has a filter, and the delay of the search bar component can now be customized in a content operation.

### DIFF
--- a/demo/tests/ReloadUITests.swift
+++ b/demo/tests/ReloadUITests.swift
@@ -1,3 +1,24 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


JohnSundell:
@spotify/objc-dev

demo/tests/ReloadUITests.swift
```diff
@@ -0,0 +1,85 @@
...
+        searchBar.typeText("s")
+        searchBar.typeText("t")
+        XCTAssertTrue(firstItem.exists)
+        XCTAssertFalse(secondItem.exists)
+        XCTAssertFalse(thirdItem.exists)
+    }
+    
+    private func addItem(named itemName: String) {
+        let app = XCUIApplication()
+        app.navigationBars["Todo List"].buttons["Add"].tap()
```
cerihughes: Should probably assert that the button exists before we tap it.

demo/tests/ReloadUITests.swift
```diff
@@ -0,0 +1,85 @@
...
+        
+        let alert = app.alerts["Add an item"]
+        let textField = alert.textFields.element(boundBy: 0)
+        textField.tap()
```
cerihughes: Assert this exists first

JohnSundell:
@cerihughes Good points, makes sense to do asserts earlier, so that it's easier to debug the tests if something goes wrong 👍 Will fix!

cerihughes:
🌳

PR #190, opened 2016-11-25
Refactor the way component wrappers are managed (+ fix memory leak)

JohnSundell:
This change refactors the way that the framework manages component wrappers, to make it more streamlined:

Current situation:
- `HUBViewController` creates component wrappers for root components itself.
- `HUBComponentReusePool` is used to create & reuse child components.

New situation:
- `HUBComponentReusePool` is used to manage all component wrappers.

While doing this, I uncovered a memory leak where header & overlay component wrappers would get retained by the framework even after they were no longer needed (this would last until their view controller would be deallocated). This has now been fixed, and a unit test for it has been added.


JohnSundell:
@spotify/objc-dev

JohnSundell:
A future commit could rename `HUBComponentReusePool` to `HUBComponentManager`.

Issue #134, opened 2016-11-09
Add custom field to the component image schema


diegc:
The component image schema should have support for custom data (type Dictionary). This enables adding per-image properties such as transformations to the images, caching policies, etc.

JohnSundell:
There's now a custom data property on `HUBComponentImageData` (implemented in https://github.com/spotify/HubFramework/pull/157), so now we need to add support for expressing it both in JSON and using `HUBComponentImageDataBuilder`.

PR #191, opened 2016-11-28
HUBComponentImageData: Finish custom data implementation

JohnSundell:
- Improve documentation
- Add builder support (both API & JSON)
- Add tests
- Add field to JSON programming guide

Resolves https://github.com/spotify/HubFramework/issues/134

### DIFF
--- a/sources/HUBComponentImageDataImplementation.m
+++ b/sources/HUBComponentImageDataImplementation.m
@@ -50,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
         _URL = [URL copy];
         _placeholderIcon = placeholderIcon;
         _localImage = localImage;
-        _customData = customData;
+        _customData = [customData copy];
     }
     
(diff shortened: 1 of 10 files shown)


JohnSundell:
@spotify/objc-dev

sources/HUBComponentImageDataImplementation.m
```diff
@@ -50,7 +50,7 @@ - (instancetype)initWithIdentifier:(nullable NSString *)identifier
         _URL = [URL copy];
         _placeholderIcon = placeholderIcon;
         _localImage = localImage;
-        _customData = customData;
+        _customData = [customData copy];
```
cerihughes: If you're `copy`ing here, should the property be marked with `copy` in the header?

JohnSundell: @cerihughes Good call, it totally should!

PR #194, opened 2016-11-29
Hiding a keyboard on drag implemented in the framework

krris:
Instead of registering a component as a `HUBComponentContentOffsetObserver` and calling `resignFirstResponder` in `updateView(forChangedContentOffset:)` we simply use `keyboardDismissMode` property which was introduced in iOS 7.


JohnSundell:
We cannot make the assumption that every app using the Hub Framework will want to have this keyboard behavior. What's wrong with implementing it in the component - isn't that a lot more explicit, and leaves more control to the app?

krris:
Right, we cannot make this assumption, my bad.

PR #197, opened 2016-11-29
Update everything for Xcode 8.1

rastersize:
Fixes a bunch of warnings and updates to Swift 3.0.1 for the demo project.

@spotify/objc-dev


spotify_os.xcconfig
```diff
@@ -62,6 +62,15 @@ HF_IGNORED_WARNINGS_0820 = $(HF_IGNORED_WARNINGS_0810)
 
 WARNING_CFLAGS = -Weverything -Wno-error=deprecated -Wno-objc-missing-property-synthesis -Wno-gnu-conditional-omitted-operand -Wno-gnu -Wno-reserved-id-macro -Wno-auto-import -Wno-missing-variable-declarations -Wno-vla -Werror $(HF_IGNORED_WARNINGS_$(XCODE_VERSION_MINOR))
 
+// Apple LLVM 8.0 - Warnings - All languages
```
8W9aG: Shouldn't these be done in the ci module?

JohnSundell: Don't think we're using it yet, right @rastersize?

rastersize: Yeah true. Will open a PR for ios-ci for those changes and then we can do bump.

rastersize: @JohnSundell You have a better memory than I do. We’re not actually using ios-ci for Hubs (more like ios-ci v2.0). So will have to update in both places.

@8W9aG I’ve sent a PR for ios-ci: https://github.com/spotify/ios-ci/pull/19

PR #198, opened 2016-11-29
View controller factory: Add simplest possible API for creating a VC

JohnSundell:
This change introduces a new convenience API for `HUBViewControllerFactory`, that enables the API user to supply only the most essential information - which content operations to use, and the title of the view controller’s feature.

The feature identifier + view URI (two other required pieces of data), are derived from the feature title.

The reasoning behind this API is to lower the barrier for entry when starting to use the framework, and to avoid having to supply non-interesting information in an app that is not URI based or doesn’t use feature identifiers in any way.


JohnSundell:
@spotify/objc-dev

include/HubFramework/HUBViewControllerFactory.h
```diff
@@ -58,12 +58,25 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
 - (nullable HUBViewController *)createViewControllerForViewURI:(NSURL *)viewURI;
 
 /**
- *  Create a view controller without a feature registration
+ *  Create a view controller without a feature registration, with implicit identifiers
+ *
+ *  @param contentOperations The content operations to use to load the content for the view controller.
+ *  @param featureTitle The title of the feature that the view controller will belong to. Used for its
+ *         default title, and also made available to contnet operations as part of `HUBFeatureInfo`.
```
rastersize: `contnet` -> `content`

JohnSundell: Ah, I've spelled this word a million times and still I get it wrong 😝 Nice catch!

PR #199, opened 2016-11-29
Introduce HUBBlockContentOperation + context

JohnSundell:
This change introduces a content operation convenience API, that enables simpler operations to be constructed using a block - `HUBBlockContentOperation`.

To avoid having to pass many parameters to the block, a contextual object API is introduced (similar to how the `HUBAction` API has a `HUBActionContext`). This contextual object API is currently only used for the newly introduced block operations, but the plan is to migrate all `HUBContentOperations` to using it, instead of having multiple parameters as input.


JohnSundell:
@spotify/objc-dev

rastersize:
Idea; how about naming the class `HUBAnonymousContentOperation`? To me that name might work nice in both Objective-C and Swift while “block” feels very Objective-C. What do you think @JohnSundell?

include/HubFramework/HUBBlockContentOperation.h
```diff
@@ -0,0 +1,51 @@
...
+@interface HUBBlockContentOperation : NSObject <HUBContentOperation>
+
+/**
+ *  Initialize an instance of this class
+ *
+ *  @param block The block to use for the operation. The block will be copied by the operation and
+ *         invoked each time the operation is performed. The contextual object passed to the block
+ *         will contain all execution parameters that are passed to any `HUBContentOperation`.
+ */
+- (instancetype)initWithBlock:(void(^)(id<HUBContentOperationContext>))block HUB_DESIGNATED_INITIALIZER;
```
rastersize: How about adding `NS_SWIFT_NAME(init(closure:))` as well?

include/HubFramework/HUBContentOperationContext.h
```diff
@@ -0,0 +1,55 @@
...
+
+@protocol HUBFeatureInfo;
+@protocol HUBViewModelBuilder;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ *  Protocol defining the public API of a content operation context
+ *
+ *  This API is currently only used with `HUBBlockContentOperation`. For more information
```
rastersize: Do we intend to send these context to “normal” content operations as well? In the future that is. In other words could we unify all operations by changing:
```objc
- (void)performForViewURI:(NSURL *)viewURI
              featureInfo:(id<HUBFeatureInfo>)featureInfo
        connectivityState:(HUBConnectivityState)connectivityState
         viewModelBuilder:(id<HUBViewModelBuilder>)viewModelBuilder
            previousError:(nullable NSError *)previousError;
```

To:
```
- (void)performForContext:(id<HUBContentOperationContext>)context;
```

---

It would have the benefit of not just unifying the API. Making sure this isn’t some special on-off edge case but a first class member. As well as making it _much_ easier to introduce, deprecate and remove “parameters” in the future.

sources/HUBBlockContentOperation.m
```diff
@@ -0,0 +1,73 @@
...
+
+#import "HUBBlockContentOperation.h"
+
+#import "HUBContentOperationContextImplementation.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface HUBBlockContentOperation ()
+
+@property (nonatomic, copy) void(^block)(id<HUBContentOperationContext>);
```
rastersize: Add `readonly`

include/HubFramework/HUBContentOperationContext.h
```diff
@@ -0,0 +1,55 @@
...
+ *  about the properties that are part of this API, refer to `HUBContentOperation`, or to
+ *  the "Content programming guide".
+ */
+@protocol HUBContentOperationContext <NSObject>
```
rastersize: Does this have to be exposed as a protocol? Seems like it’s just a “model” object and could as such, just as easily, be a readonly class. Benefit would that we’d have less symbols and code.

include/HubFramework/HUBContentOperationContext.h
JohnSundell: Yeah, I wrote this in the PR description 😉 

> This contextual object API is currently only used for the newly introduced block operations, but the plan is to migrate all HUBContentOperations to using it, instead of having multiple parameters as input.

sources/HUBBlockContentOperation.m
JohnSundell: Do'h. What a rookie mistake! 😅

include/HubFramework/HUBBlockContentOperation.h
JohnSundell: Totally missed that, thanks 👍

include/HubFramework/HUBContentOperationContext.h
JohnSundell: Actually, I think I prefer to keep it the way it is currently implemented. The reason is that it creates a very clear separation between the API and the implementation. It clearly signals to the API user that this is not a class that you are supposed to instantiate yourself - since the initialzier is not accessible.

include/HubFramework/HUBContentOperationContext.h
rastersize: Ooooooh, haha. Well sorry about that! I should work on my readying skills 😄

include/HubFramework/HUBContentOperationContext.h
rastersize: Alright that’s a pretty good reason! 👍

Issue #163, opened 2016-11-16
Live: Reset ANSI color output after printing error message


JohnSundell:
If the `hublive` command line interface encounters an error, it prints that error using the red ANSI color. However, it doesn't restore the original text color of the user's terminal, meaning that all subsequent messages will be of the same red error color.

Here's a screenshot reported by @josete89 in the Slack channel:
<img width="1550" alt="screen_shot_2016-11-12_at_10 31 24_am" src="https://cloud.githubusercontent.com/assets/2466701/20355622/38dba972-ac22-11e6-9588-ba045e939281.png">

We should fix this, and make sure that the terminal color is always restored after an error message is printed.

This is the code used to print with a color: https://github.com/spotify/HubFramework/blob/master/live/sources/String%2BPrintWithColor.swift

rastersize:
Fixed with #200.

PR #200, opened 2016-11-30
Fix colored strings to restore original terminal color

mpluders:
A potential fix for #163. It simply appends a white Ansi color code when printing colored strings. This successfully restores the original terminal color.

I'm not sure how you would create a unit test for something like this, but would love to hear any suggestions.

### DIFF
--- a/live/sources/String+PrintWithColor.swift
+++ b/live/sources/String+PrintWithColor.swift
@@ -29,6 +29,7 @@ extension String {
      *  - Parameter color: The color to print the string using
      */
     func print(withColor color: AnsiColor) {
-        Swift.print(color.rawValue + self)
+        /// Append a reset color code to the string, to clear all colors and styles
+        Swift.print(color.rawValue + self + "\u{001B}[0;0m")
     }
 }


live/sources/String+PrintWithColor.swift
```diff
@@ -29,6 +29,7 @@ extension String {
      *  - Parameter color: The color to print the string using
      */
     func print(withColor color: AnsiColor) {
-        Swift.print(color.rawValue + self)
+        /// Append a white color code to the string, to reset the console color
+        Swift.print(color.rawValue + self + AnsiColor.white.rawValue)
```
rastersize: How about using the “reset” ANSI value instead? Setting it to white might cause issues for user’s that use a white background color and black foreground. The “reset” value instead just clears any custom formatting. Returning the text that comes afterwards to the terminals default colors.

The ansi value is:
`[0m`

So we’d add a enum value like:
```swift
case reset = "\u{001B}[0m"
```

JohnSundell: Ah, that sounds a lot better! 👍 We actually don't need a case for it, we can just inline the string at the end of the `print` call.

mpluders:
Ahh that makes more sense, thanks for point that out @rastersize. I implemented the reset code in-line, and removed the superfluous case from AnsiColor.swift.

JohnSundell:
Awesome @mpluders! Looking great 👍

PR #202, opened 2016-11-30
Notify HubContentOffsetObserver on the first appearance.

krris:
This pull request fixes the following bug: `HubContentOffsetObserver` in not notified on the first appearance.

To wrap up the behaviour of `HubContentOffsetObserver`:
* it should be notified on the first appearance,
* it should be notified if `contentOffset` has changed,
* it shouldn't be notified if a `viewController` has been reloaded.

If you have any suggestions how it can be improved or if I forgot about any test case, please let me know 💡


sources/HUBComponentWrapper.h
```diff
@@ -167,6 +167,9 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState;
 /// Whether the wrapped component's view has appeared since the model was last changed
 @property (nonatomic, readonly) BOOL viewHasAppearedSinceLastModelChange;
 
+/// Whether the wrapped component's view has appeared for the first time
+@property (nonatomic, assign, nullable, readonly) NSNumber *viewHasAppearedForTheFirstTime;
```
JohnSundell: Why is it an `NSNumber` and not simply a `BOOL`? (also, how about shortening it down to `viewHasAppeared` - since `ForTheFirstTime` is implied).

sources/HUBComponentWrapper.m
```diff
@@ -551,6 +554,15 @@ - (void)updateViewForSelectionState:(HUBComponentSelectionState)selectionState n
     }
 }
 
+- (void)configureViewHasAppearedForTheFirstTime
+{
+    if ([self.viewHasAppearedForTheFirstTime boolValue] == YES) {
```
JohnSundell: Can't this simply be replaced with `self.viewHasAppeared = YES`?

sources/HUBComponentWrapper.h
krris: Because it can have 3 states:
* `nil` - `viewWillAppear` hasn't been called,
* `true` - `viewWillAppear` has been called only once,
* `false` - `viewWillAppear` has been called more than 1 time.

It's needed for example when a content operation is rescheduled. In the situation, when the component has appeared before and `contentOffset` hasn't been changed, we don't want to notify a `contentOffsetObserver`. We want to only notify it on the first appearance.

sources/HUBComponentWrapper.m
krris: IMO it doesn't describe well this property. See my previous comment.

sources/HUBComponentWrapper.h
JohnSundell: How about using an integer to keep track of the apperance count instead? More explicit than relying on `nil` semantics (IMO).

JohnSundell: Alternatively, an enum:
```
enum ApperanceState {
    case hasNotAppeared
    case hasAppearedOnce
    case hasAppearedMultipleTimes
}
```

Personally, I think I prefer the integer approach (if possible). Simpler 😄

krris: After writing this comment I got the same idea 😄  Will do this 👍

JohnSundell: 👍

krris:
@JohnSundell 🎾

sources/HUBComponentWrapper.h
```diff
@@ -167,6 +167,9 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState;
 /// Whether the wrapped component's view has appeared since the model was last changed
 @property (nonatomic, readonly) BOOL viewHasAppearedSinceLastModelChange;
 
+/// The wrapped component's appearance count
```
JohnSundell: Maybe explain a bit more what this actually does? So it's not just simply re-stating the property name 😅 How about:
```
/**
 *    The number of times the wrapped component has appeared on the screen
 *
 *    Incremented every time the component gets sent the -viewWillAppear message.
 */
```

krris: Yes, yes, yes 😄

PR #203, opened 2016-12-01
JSON schema registry: Add API for unregistering a custom schema

JohnSundell:
Was not yet implemented for some reason. All other registries offer an unregistration API, so makes sense for the JSON schema one to do so too.


JohnSundell:
@spotify/objc-dev

PR #205, opened 2016-12-01
Fix crash when overlapping render requests are made

cerihughes:
We were seeing a crash where a render request was made while an existing render was in progress. This resulted in a new view model being set while the collection view was being updated, which in turn resulted in a 💥 because the number of items in the model wasn't consistent with the items in the collection view at the end of the render.

This PR:
- Updates the `HUBViewController` to allow us to inject the `HUBViewModelRenderer` (makes testing easier),
- Passes the collection view into the render method instead of injecting it on creation,
- Stores "pending" view models in the VC, and only assigns them when any previous render operation has finished,
- Adds tests for the crash case.


tests/HUBViewControllerTests.m
```diff
@@ -2866,6 +2874,68 @@ - (void)testThatDelegateIsNotifiedWhenHeaderAppears
...
+    HUBViewModelRendererMock *viewModelRenderer = [HUBViewModelRendererMock new];
+    [self createViewControllerWithViewModelRenderer:viewModelRenderer];
+
+    self.contentOperation.contentLoadingBlock = ^(id<HUBViewModelBuilder> viewModelBuilder) {
+        [viewModelBuilder headerComponentModelBuilder].title = @"title1";
+        return YES;
+    };
+
+    [self.viewController.view layoutIfNeeded];
+    [self.viewController reload];
```
JohnSundell: Instead of this, how about calling `[self simulateViewControllerLayoutCycle]`? (to be consistent with the other tests)

tests/mocks/HUBViewModelRendererMock.h
```diff
@@ -0,0 +1,34 @@
...
+
+#import "HUBViewModelRenderer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Mocked renderer that gives the test control over when to fire the completion block. For use in tests only.
+@interface HUBViewModelRendererMock : HUBViewModelRenderer
+
+/// All completion blocks passed to the render method
+@property (nonatomic, readonly) NSMutableArray<void(^)(void)> *completionBlocks;
```
JohnSundell: `strong`?

tests/HUBViewControllerTests.m
```diff
@@ -2866,6 +2874,68 @@ - (void)testThatDelegateIsNotifiedWhenHeaderAppears
...
+{
+    HUBViewModelRendererMock *viewModelRenderer = [HUBViewModelRendererMock new];
+    [self createViewControllerWithViewModelRenderer:viewModelRenderer];
+
+    self.contentOperation.contentLoadingBlock = ^(id<HUBViewModelBuilder> viewModelBuilder) {
+        [viewModelBuilder headerComponentModelBuilder].title = @"title1";
+        return YES;
+    };
+
+    [self.viewController.view layoutIfNeeded];
```
JohnSundell: The above comment also applies here 🙂

cerihughes:
@JohnSundell - thanks for the comments! Updated...

JohnSundell:
Built and verified locally ✅

PR #206, opened 2016-12-01
Fix issues with components reuse

mvlasieva:
This PR fixes a couple of issues with reusing both parent and child components. In makes sure components and collection view cells can be reused independently and cells in the collection view cache which are still holding a reference to already reused components don't interfere with them.
@JohnSundell


sources/HUBViewController.m
```diff
@@ -708,19 +706,16 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
...
-        [self didAddComponentWrapper:componentWrapper];
-    }
-    
-    HUBComponentWrapper * const componentWrapper = [self componentWrapperFromCell:cell];
+
+    HUBComponentWrapper * const componentWrapper = [self.componentReusePool componentWrapperForModel:componentModel
+                                                                                            delegate:self
+                                                                                              parent:nil];
+
+    self.componentWrappersByCellIdentifier[cell.identifier] = componentWrapper;
```
JohnSundell: Are we ever clearing this up? That is, before we assign a new wrapper to the cell, we should do:
```
self.componentWrappersByCellIdentifier[cell.identifier] = nil;
```

mvlasieva: Sorry, I don't get it. What is the point of removing object from the dictionary right before setting a new object for the same key?

JohnSundell: You are completely right, sorry - I was confused 😆

JohnSundell:
@mvlasieva Since Travis are currently undergoing an incident, I'm using my computer as our build server 😅 However, this PR doesn't pass the tests locally. Can you update the PR so that they do? 🙂

mvlasieva:
@JohnSundell All done, feel free to review.

PR #208, opened 2016-12-01
Adds a new diffing algorithm that performs better on larger, homogenous datasets

Goos:
Also adds an option to perform diffing using different algorithms.

This new algorithm has a complexity of O((N + M) * D), where N and M is the two dataset lengths, and D is the minimum number of changes between them, whereas the previous algorithm had O(N * M). 

This means the new algorithm _should_ be faster in most normal cases, such paginated loading where much of the content remains unchanged.


sources/HUBViewModelDiff.h
```diff
@@ -24,6 +24,29 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class HUBViewModelDiff;
+
+typedef HUBViewModelDiff *(HUBDiffAlgorithm)(id<HUBViewModel>, id<HUBViewModel>);
```
rastersize: Document please

rastersize: How come this is implemented as (_long_) C functions? Did you consider making this into a protocol instead? Such as:
```objc
@protocol HUBDiffAlgorithm
- (HUBViewModelDiff *)differenceGoingFromViewModel:(id<HUBViewModel>)fromViewModel
                                       toViewModel:(id<HUBViewModel>)toViewModel;
@end
```

It would, imo, allow for more flexibility. For example a `debugName` property could be added allowing the `-debugDescription` method further down to print which algorithm is used. It might also make the code feel less scary for non-C developers.

Not opposed to using passing function pointers like this. Just interested in your reasoning @Goos.

sources/HUBViewModelDiff.m
```diff
@@ -52,24 +52,42 @@ - (instancetype)initWithInserts:(NSIndexSet *)inserts
     return self;
 }
 
-+ (NSArray<NSString *> *)componentIdentifiersFromViewModel:(id<HUBViewModel>)viewModel
++ (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
+                      toViewModel:(id<HUBViewModel>)toViewModel
+                        algorithm:(HUBDiffAlgorithm)algorithm
+{
+    return algorithm(fromViewModel, toViewModel);
```
rastersize: How about asserting that the function pointer isn’t `NULL` first?

sources/HUBViewModelDiff.m
```diff
@@ -145,15 +163,234 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
...
+        
+        return [[self alloc] initWithFromPoint:HUBDiffPointMake(x, x - step.k - 1) toPoint:HUBDiffPointMake(x, x - step.k) D:step.D];
+    } else {
+        NSInteger x = step.previousX + 1;
+
+        return [[self alloc] initWithFromPoint:HUBDiffPointMake(x - 1, x - step.k) toPoint:HUBDiffPointMake(x, x - step.k) D:step.D];
+    }
+}
+
+- (instancetype)initWithFromPoint:(HUBDiffPoint)fromPoint toPoint:(HUBDiffPoint)toPoint D:(NSInteger)D
```
rastersize: How about naming the `D` parameter `minimumNumberOfChanges` instead? Since that’s what it represents if I understand your API docs further up correctly? Would make the code a bit more readable 😄

sources/HUBViewModelDiff.m
```diff
@@ -145,15 +163,234 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
...
+    NSInteger x;
+    NSInteger y;
+} HUBDiffPoint;
+
+static inline HUBDiffPoint HUBDiffPointMake(NSInteger x, NSInteger y) {
+    return (HUBDiffPoint){ .x = x, .y = y};
+}
+
+typedef struct {
+    NSInteger D;
```
rastersize: How about naming it `minimumNumberOfChanges`? Makes it a bit more readable 😸

sources/HUBViewModelDiff.m
```diff
@@ -145,15 +163,234 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
...
+    HUBDiffTraceTypeInsert,
+    HUBDiffTraceTypeDelete,
+    HUBDiffTraceTypeMatchPoint
+};
+
+@interface HUBDiffTrace : NSObject
+
+@property (nonatomic, assign, readonly) HUBDiffPoint from;
+@property (nonatomic, assign, readonly) HUBDiffPoint to;
+@property (nonatomic, assign, readonly) NSInteger D;
```
rastersize: `minimumNumberOfChanges`?

sources/HUBViewModelDiff.m
```diff
@@ -145,15 +163,234 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
...
+    } else if (fromViewModel.bodyComponentModels.count == 0) {
+        return  HUBDiffInsertionTracesFromViewModel(toViewModel);
+    } else if (toViewModel.bodyComponentModels.count == 0) {
+        return HUBDiffDeletionTracesFromViewModel(fromViewModel);
+    } else {
+        return HUBDiffMyersTracesBetweenViewModels(fromViewModel, toViewModel);
+    }
+}
+
+static NSArray<HUBDiffTrace *> *HUBDiffFindPathFromTraces(NSArray *traces) {
```
rastersize: Can we make the argument have the type `NSArray<HUBDiffTrace *> *`?

sources/HUBViewModelDiff.h
```diff
@@ -24,6 +24,33 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class HUBViewModelDiff;
+
+/**
+ * The prototype of a function used to calculate a list of changes to get from
+ * the one view model to another.
```
JohnSundell: How about documenting the parameters, so it's clear which one is the from & to view models?

sources/HUBViewModelDiff.m
```diff
@@ -145,15 +164,237 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
...
 }
 
-- (NSString *)debugDescription
+typedef struct {
+    NSInteger x;
+    NSInteger y;
+} HUBDiffPoint;
+
+static inline HUBDiffPoint HUBDiffPointMake(NSInteger x, NSInteger y) {
+    return (HUBDiffPoint){ .x = x, .y = y};
```
JohnSundell: 🎨 Style is a bit inconsitent here, you have a leading space but not a trailing one 😄

sources/HUBViewModelDiff.m
```diff
@@ -145,15 +164,237 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
...
+    HUBDiffTraceTypeDelete,
+    HUBDiffTraceTypeMatchPoint
+};
+
+@interface HUBDiffTrace : NSObject
+
+@property (nonatomic, assign, readonly) HUBDiffPoint from;
+@property (nonatomic, assign, readonly) HUBDiffPoint to;
+@property (nonatomic, assign, readonly) NSInteger changes;
+
```
JohnSundell: Why this empty row? 😱

sources/HUBViewModelDiff.m
```diff
@@ -145,15 +164,237 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
...
+        
+        return [[self alloc] initWithFromPoint:HUBDiffPointMake(x, x - step.diagonal - 1) toPoint:HUBDiffPointMake(x, x - step.diagonal) changes:step.changes];
+    } else {
+        NSInteger x = step.previousX + 1;
+
+        return [[self alloc] initWithFromPoint:HUBDiffPointMake(x - 1, x - step.diagonal) toPoint:HUBDiffPointMake(x, x - step.diagonal) changes:step.changes];
+    }
+}
+
+- (instancetype)initWithFromPoint:(HUBDiffPoint)fromPoint toPoint:(HUBDiffPoint)toPoint changes:(NSInteger)changes
```
JohnSundell: What about `type`?

sources/HUBViewModelDiff.h
Goos: added below!

Goos:
@JohnSundell Got it! More documentation coming up

JohnSundell:
@Goos Awesome stuff! 🚀

PR #209, opened 2016-12-01
Fix issues with building HubFramework in certain environments

rastersize:
- Removes all extraneous overrides of build settings from the HubFramework library target.
  - These are covered by our `spotify_os.xcconfig` file, so that we can override them internally.
  - Some didn’t have any effect on the type of target so I removed them.
- Add two files which where missing from the umbrella header. Leading to warnings for users that imported the library as a clang module (e.g. apps in Swift, like the demo).

@spotify/objc-dev


JohnSundell:
✅ Built and verified locally

PR #211, opened 2016-12-04
Cast to CGFloat to fix 32bit builds.

cerihughes:
The build failed locally on iPhone4s (for example)


cerihughes:
@spotify/objc-dev master is currently "broken" (not compiling on 32bit). This fixes it.

PR #212, opened 2016-12-04
Templates: Remove viewURI argument from HUBContentOperationActionObserver

JohnSundell:
Argument was removed a while back, but we missed to update the template.


JohnSundell:
@spotify/objc-dev

PR #213, opened 2016-12-04
Add constant for default component namespace + test it

JohnSundell:
This change moves what was previously an inline string for the default component namespace (used when setting up the Hub Framework using the simplest convenience API) to a constant. It also adds a test to make sure that the constant is really used.


JohnSundell:
@spotify/objc-dev

PR #215, opened 2016-12-05
Setup guide: Recommend the simplest way of setting up HUBManager

JohnSundell:
The setup guide has been updated to recommend that the simplest initializer is used to setup `HUBManager`.

### DIFF
--- a/documentation/Setup guide.md
+++ b/documentation/Setup guide.md
@@ -28,12 +28,12 @@ The Hub Framework requires you to link with Apple's `SystemConfiguration` framew
 
 Each application using the Hub Framework will have one instance of the `HUBManager` class, which manages an instance of the framework. It’s recommended that you create this object early in your application’s lifecycle and keep it somewhere accessible - for example in your `AppDelegate`.
 
-In order to create a `HUBManager` instance, you need to supply implementations of the following 2 protocols:
+There are a few different ways to setup `HUBManager`, each providing different levels of customizability. The simplest way is to use the defaults, and just supply the following two parameters:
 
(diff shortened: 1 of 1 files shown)


documentation/Setup guide.md
```diff
@@ -28,12 +28,12 @@ The Hub Framework requires you to link with Apple's `SystemConfiguration` framew
 
 Each application using the Hub Framework will have one instance of the `HUBManager` class, which manages an instance of the framework. It’s recommended that you create this object early in your application’s lifecycle and keep it somewhere accessible - for example in your `AppDelegate`.
 
-In order to create a `HUBManager` instance, you need to supply implementations of the following 2 protocols:
+There are a few different ways to setup `HUBManager`, each providing different levels of customizability. The simplest way is to use the defaults, and just supply the following two parameters:
 
-- `HUBComponentLayoutManager`: For calculating margins and other layout attributes for components based on their layout traits.
-- `HUBComponentFallbackHandler`: For providing fallback information in case a component model’s component couldn’t be resolved by the framework.
+- `componentMargin`: What margin that should be used in between components.
+- `componentFallbackBlock/Closure`: A block/closure that returns a component to use for fallback, in case a default one couldn't be matched.
```
Goos: Perhaps 
> in case a default one couldn't be matched

could be clarified with something like

> in case a component identifier with no matching component factory is provided

JohnSundell: Yeah, true. Wanted to keep things short and sweet here though, since the other guides (and the documentation for components & their factories) go into a lot more detail 😄

PR #216, opened 2016-12-05
Automatically manage selection for child components

JohnSundell:
This change makes the Hub Framework assume the responsibility of handling selection for child components, when those components were created through the framework (and not by using a custom `UIView`).

It’s become clear that it’s ambiguous (and very cumbersome) for each component to have to manually handle selection for its children. As long as we were mostly implementing containers (such as Carousels) as components with children, this was fine, as we could leverage nested `UICollectionViews` and `HUBComponentCollectionViewCell`. However, as we are creating more and more complex component hierarchies, with containers that add backgrounds and other fancy stuff - it started to become a problem.

So, with this change, parent components are no longer responsible for the selection of their children. The `component(_:childSelectedAt:customData:)` delegate method has been replaced with `component(_:childWithCustomViewSelectedAt:customData:)`, to clearly signal that this method is only meant to be called for custom `UIView` children.

The implementation is using the same `HUBComponentGestureRecognizer` as for root components to handle selection, and it now also keeps track of the frame of a component, so that nested scrolling (such as in a carousel) can be handled in the correct way.

### DIFF
--- a/demo/tests/SelectionUITests.swift
+++ b/demo/tests/SelectionUITests.swift
@@ -66,12 +66,12 @@ class SelectionUITests: UITestCase {
         XCTAssertTrue(rootCell.exists)
         
         let carouselCellA = rootCell.cells.element(boundBy: 0)
-        let carouselCellB = rootCell.cells.element(boundBy: 1)
+        let carouselCellB = rootCell.cells.element(boundBy: 2)
         XCTAssertTrue(carouselCellA.exists)
         XCTAssertTrue(carouselCellB.exists)
         
         // Start pressing, then move to the next cell. No selection should happen = we're still in pretty pictures
-        carouselCellA.press(forDuration: 1, thenDragTo: carouselCellB)
+        carouselCellA.press(forDuration: 3, thenDragTo: carouselCellB)
         XCTAssertTrue(app.navigationBars["Pretty Pictures"].exists)
     }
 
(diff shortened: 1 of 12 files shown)


JohnSundell:
@spotify/objc-dev

demo/tests/SelectionUITests.swift
```diff
@@ -56,6 +54,26 @@ class SelectionUITests: UITestCase {
...
+        
+        let rootCell = collectionView.children(matching: .cell).element(boundBy: 0)
+        XCTAssertTrue(rootCell.exists)
+        
+        let carouselCellA = rootCell.cells.element(boundBy: 0)
+        let carouselCellB = rootCell.cells.element(boundBy: 1)
+        XCTAssertTrue(carouselCellA.exists)
+        XCTAssertTrue(carouselCellB.exists)
+        
+        // Start pressing, then move to the next cell. No selection should happen = we're still in pretty pictures
```
krris: Nice test case 👍

krris:
@JohnSundell Some tests are failing. If I run them locally it's fine, but Travis is screaming about ` testChildSelectionCancelledOnParentScrolling, XCTAssertTrue failed -`.

JohnSundell:
@krris Yeah, I'm gonna add some longer timeouts to it - I've seen this happen to other UI tests as well when run on Travis.

krris:
Perfect! 👏

PR #218, opened 2016-12-06
Add ability to perform actions from scroll handler

mvlasieva:
This covers a use case when actions need to be performed in relation to scrolling in Hub view controller (for example centering new item).


JohnSundell:
I'm not sure if this is such a good idea - it gives the scroll handler a bit too many responsibilities. Is there no way to perform these actions from a content operation or component instead?

mvlasieva:
The only object that is notified of scrollingDidEnd events is the scroll handler. Scroll handler has no way to access any component, so it can't be the component who performs the action. Performing this action from content operation seems completely irrelevant.
Another approach would be to notify components conforming to HUBComponentContentOffsetObserver of scrollingDidEnd events in addition to scrollViewDidScroll. Would it be a better solution?

PR #219, opened 2016-12-06
Add dynamic framework target + use it in the demo app

JohnSundell:
This will enable us to support Carthage is the proper way. The demo app has also been updated to use the dynamic framework instead of the static one, and to run tests against the dynamic framework. This will give us a fairly good safety net that the static library and dynamic framework will maintain feature pairity.

To enable tests to be run against the dynamic framework, minor changes needed to be made to the code:

- Use `XCTAssertEqualObjects` instead of `XCTAssertEqual` to compare
  string pointers.
- Enable a bundle to be injected when loading local images, since a
  dynamic framework and its test suite do not share the same bundle.

Resolves https://github.com/spotify/HubFramework/issues/42


JohnSundell:
@spotify/objc-dev

mhallendal:
👍

sources/Info.plist
```diff
@@ -0,0 +1,24 @@
...
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
```
8W9aG: Should we make this the current version?

JohnSundell: Yeah that's a good idea. Is there a magic variable for that?

8W9aG: Unfortunately I don't think so, in our other frameworks its set manually, but I could absolutely be wrong

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -1473,6 +2161,121 @@
 			};
 			name = Release;
 		};
+		8AE6C0051DF6E3110063B2B1 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
```
rastersize: All of these should preferably live in the, or another, xcconfig file. Since our spotify_os file is a bit… unwieldy… we can do this change in a subsequent PR. It would basically be nice to split it up. Or use a [better set of configs](https://github.com/jspahrsummers/xcconfigs) for the framework, live and demo targets. I’ll take that task.

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -1473,6 +2161,121 @@
...
+				COPY_PHASE_STRIP = NO;
+				CURRENT_PROJECT_VERSION = 1;
+				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEFINES_MODULE = YES;
+				DYLIB_COMPATIBILITY_VERSION = 1;
+				DYLIB_CURRENT_VERSION = 1;
+				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				ENABLE_TESTABILITY = YES;
+				GCC_C_LANGUAGE_STANDARD = gnu99;
```
rastersize: We use `gnu11`.

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -1473,6 +2161,121 @@
...
+				);
+				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
+				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
+				GCC_WARN_UNDECLARED_SELECTOR = YES;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				GCC_WARN_UNUSED_FUNCTION = YES;
+				GCC_WARN_UNUSED_VARIABLE = YES;
+				INFOPLIST_FILE = "$(SRCROOT)/sources/Info.plist";
+				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
+				IPHONEOS_DEPLOYMENT_TARGET = 10.1;
```
rastersize: The deployment target should be iOS 8.

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -1473,6 +2161,121 @@
...
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				ONLY_ACTIVE_ARCH = YES;
+				PRODUCT_BUNDLE_IDENTIFIER = "com.spotify.HubFramework-iOS";
+				PRODUCT_NAME = HubFramework;
+				SKIP_INSTALL = YES;
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 3.0;
```
rastersize: Change to `3.0.1`, otherwise we get “use legacy version”

sources/HUBComponentImageDataBuilderImplementation.h
```diff
@@ -31,6 +31,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// Concrete implementation of the `HUBComponentImageDataBuilder` API
 @interface HUBComponentImageDataBuilderImplementation : NSObject <HUBComponentImageDataBuilder, NSCopying>
 
+/// Any specific bundle that the builder should use to load local images (defaults to the main bundle)
+@property (nonatomic, weak, nullable) NSBundle *bundle;
```
rastersize: Any reason this is `weak`?

tests/HUBComponentImageDataBuilderTests.m
```diff
@@ -49,6 +49,7 @@ - (void)setUp
                                                                                       iconImageResolver:iconImageResolver];
     
     self.builder = [[HUBComponentImageDataBuilderImplementation alloc] initWithJSONSchema:JSONSchema iconImageResolver:iconImageResolver];
+    self.builder.bundle = [NSBundle bundleForClass:[self class]];
```
rastersize: Nit: `class` is a property on `NSObject` instances. As such dot-notation should be used according to our style guide.

sources/HUBComponentImageDataBuilderImplementation.h
JohnSundell: Didn't feel right for the image data builder to retain the bundle, since it shouldn't assume any ownership over it.

JohnSundell:
@rastersize Updated 🚀

sources/HUBComponentImageDataBuilderImplementation.h
rastersize: Well that means the user of the framework will have to retain the `NSBundle` object for the framework. Which makes it a bit more hairy and tricky to use. Since there’s no guarantee that the system will retain it for you.

rastersize: @JohnSundell ☝️

rastersize: @JohnSundell Okay.

PR #220, opened 2016-12-06
Fix importing HubFramework if other lib imports it

rastersize:
The modulemap should by default, for Xcode to not have a hard time, be named exactly `module.modulemap`. As well as live in the same directory as the public headers.

This fixes importing HubFramework as a clang module if some other clang module also imports it. Which is the case with, for example, a component library.

@spotify/objc-dev


JohnSundell:
👍

mumme:
👍

PR #221, opened 2016-12-06
Fix and improve CI scripts

rastersize:
This PR fixes a few and improves a few things with our CI scripts.

### Changes
- Fix UI tests overwriting the logs from the unit tests.
- Fix code coverage being ran twice.
  - Now it’ll run only run as the last step of `ci_all`, iff we should run any type of tests.
- Remove redundant recreation of demo Xcode scheme.
  - It’s already shared.
- Update Fastlane to 1.111.x.
- Fix demo app and unit tests being built and ran twice each time on CI.
  - This happened with #219 (add dynamic framework PR).
  - The demo app’s Xcode scheme contains both the UI and unit tests for the test action.
  - If the UI tests should be ran we skip both the `RUN_TESTS` and `BUILD_DEMO_APP` steps since they’re implicit.


rastersize:
Hmm, this didn’t fix the duplicate unit test run 😞 At least it made Danger report both of them, hehe 😬

rastersize:
This is ready for review @spotify/objc-dev

It should reduce time spent in CI as well as improve our flakiness situation. Since we won’t re-build the world and re-run the unit tests at least two times anymore 🎉🌮

PR #223, opened 2016-12-07
View controller: Make view URI a public property

JohnSundell:
No good reason to keep it private, and it makes some operations that handle view URIs a lot simpler, since they can just access it from the view controller directly, instead of having to pass another parameter around.


JohnSundell:
@spotify/objc-dev

PR #224, opened 2016-12-07
HUBComponentWithSelectionState: Remove doc comment about cells

JohnSundell:
Cells have automatic selection state handling, so no need to use the protocol for those.


JohnSundell:
@spotify/objc-dev

PR #226, opened 2016-12-08
Add test for latest crash

cerihughes:
Testing this proved more difficult than expected due to the fact that the `UICollectionView` kept triggering a lot of internal logic that required the setup of all of its data sources, reuse identifiers, layout traits etc. I found a way to silence this by overriding the batch update methods.

The test calls 2 render operations - 1 from inside the completion block of the other. Before the fix, the "last rendered" view model wouldn't have been cached at this point, so both render operations would have resulted in `nil` diffs being generated.

After the fix, the "last rendered" view model is set before the completion block is called, so the 2nd render operation should generate a diff. The test captures all diffs sent to the layout, and asserts on them.

The PR also:

- shares the view model utility code created by the diffing tests.
- creates mocks for the layout and component registry.
- makes sure the collection view is using a mock layout, not a standard `UICollectionViewLayout` (or it crashes when rendering).


cerihughes:
@JohnSundell @rastersize

tests/HUBViewModelRendererTests.m
```diff
@@ -0,0 +1,126 @@
...
+
+    [self waitForExpectationsWithTimeout:2 handler:nil];
+
+    XCTAssertEqual(self.collectionViewLayout.capturedViewModels.count, 2u);
+    XCTAssertEqualObjects(self.collectionViewLayout.capturedViewModels[0], firstViewModel);
+    XCTAssertEqualObjects(self.collectionViewLayout.capturedViewModels[1], secondViewModel);
+
+    XCTAssertEqual(self.collectionViewLayout.capturedViewModelDiffs.count, 2u);
+    // The first invocation shouldn't generate a diff.
+    XCTAssertEqualObjects(self.collectionViewLayout.capturedViewModelDiffs[0], [NSNull null]);
```
cerihughes: @rastersize That's a good point, and most likely the reason why the builds are crashing on Travis. I'll fix it.

cerihughes: (although, I'm not sure why the test runs differently on the server and produces different results)

cerihughes:
@rastersize Made the changes you suggested and also fixed the crash (I suspect it was caused by https://github.com/spotify/HubFramework/pull/234)

PR #227, opened 2016-12-08
Memory management in tests

cerihughes:
We need to make sure we `nil` all of our strong properties in the `tearDown` phase of unit tests.

*Errrr... Why?*
`XCTest` will create a new instance of your `XCTestCase` subclass for each test method in that class, and it will *not* release that test case for the lifetime of the tests being run. Maybe that's a bug in `XCTest`, or maybe it's by design. I'm not sure, but it's the way it's always been.

This means that any properties created during a test will never be released for us - `tearDown` is the only real opportunity we have to step in and release references to these instances.

*But why do we care during testing?*
- We want to test our `dealloc` methods too, right?
- Profiling unit tests is much more difficult when every property created in every test is still alive.
- If we ever introduce code that sends `NSNotification`s (for example), we'll potentially have problems where old instances could be producing / consuming events that we don't expect.

### DIFF
--- a/tests/HUBActionRegistryTests.m
+++ b/tests/HUBActionRegistryTests.m
@@ -42,6 +42,12 @@
     self.actionRegistry = [HUBActionRegistryImplementation registryWithDefaultSelectionAction];
 }
 
+- (void)tearDown
+{
+    self.actionRegistry = nil;
+    [super tearDown];
+}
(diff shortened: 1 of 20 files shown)


cerihughes:
@spotify/objc-dev

PR #228, opened 2016-12-11
Add ability to specify keyboardDismissMode by the scroll handler

nataliq:
`UIScrollView`'s `keyboardDismissMode` is available since iOS 7 and is the preferred way for configuring the automatic keyboard dismissal. This option can reduce a lot of custom content offset observing code so less bugs and code duplication.

### DIFF
--- a/include/HubFramework/HUBViewControllerScrollHandler.h
+++ b/include/HubFramework/HUBViewControllerScrollHandler.h
@@ -55,7 +55,7 @@
 - (BOOL)shouldAutomaticallyAdjustContentInsetsInViewController:(HUBViewController *)viewController;
 
 /**
- *  Return the manner in which the keyboard is dismissed when a drag begins in the view controller
+ *  Return the manner in which the keyboard is dismissed when the content in the view controller is dragged
  *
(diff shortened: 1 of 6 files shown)


include/HubFramework/HUBViewControllerScrollHandler.h
```diff
@@ -55,6 +55,17 @@
...
 /**
+ *  Return the manner in which the keyboard is dismissed when a drag begins in the view controller
+ *
+ *  @param viewController The view controller in question
+ *
+ *  The Hub Framework will call this method when a view controller is being set up. The returned value will be
+ *  assigned to the `keyboardDismissMode` property of the used scroll view, so see the documentation for that property
+ *  on `UIScrollView` for more information.
+ */
+- (UIScrollViewKeyboardDismissMode)keyboardDismissModeForDraggingViewController:(HUBViewController *)viewController;
```
JohnSundell: Since this is only called once, I think it makes more sense to name this method `keyboardDismissModeForViewController:`. With the current naming, it sounds like it will be called once dragging starts. Also, the view controller is not really being dragged 😄

PR #229, opened 2016-12-11
Add proposed value for the scroll handler to be used by features

nataliq:
This PR moves the logic from the `HUBViewControllerDefaultScrollHandler` to the `HUBViewController` so that it could be reused by features that want to provide custom scroll handler to override some behaviour but don't want to replicate this logic.

### DIFF
--- a/include/HubFramework/HUBViewControllerScrollHandler.h
+++ b/include/HubFramework/HUBViewControllerScrollHandler.h
@@ -114,16 +114,18 @@
 /**
  *  Return the content offset for displaying a component at a certain scroll position.
  *  
- *  @param componentIndex The index of the component to display.
- *  @param scrollPosition The position to display the component at.
+ *  @param componentIndex The index of the component to display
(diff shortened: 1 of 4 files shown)


JohnSundell:
@Goos I think it'd be great if you reviewed this, since you've been working a lot on the scrolling behaviors 😄

Goos:
@JohnSundell, @nataliq I'm kind of hesitant to move back scrolling behaviour to the view controller as it's way too large already. How about exposing the HUBDefaultScrollHandler as part of the interface for subclassing instead?

nataliq:
@Goos I like that idea. Also currently if you want to provide different value for 1-2 things you still have to implement all methods..

JohnSundell:
@Goos @nataliq Awesome idea! We should probably do that with more of our default classes 👍

nataliq:
@Goos @JohnSundell 
Here we go: https://github.com/spotify/HubFramework/pull/230

PR #230, opened 2016-12-12
Expose the default scroll handler as a override point

nataliq:
Moves the `HUBViewControllerDefaultScrollHandler` to the include folder so that it could be subclassed by features that want to provide different implementation just for few methods. Apart from that only the description of the class is slightly modified.


nataliq:
@Goos right, added. It was a small change so I rebased hope you don't mind :D

Goos:
No worries, still approved :smile: 👍

rastersize:
👍

rastersize:
Closing this PR in favor of #250 which takes @nataliq’s commits and then also applies a final commit to fix the build error.

PR #231, opened 2016-12-12
Add feature info to HUBActionContext

JohnSundell:
This change makes `HUBActionContext` carry a `HUBFeatureInfo` object that contains information about the feature that the action is being performed in. This was previously only exposed to content operations.


JohnSundell:
@spotify/objc-dev

PR #237, opened 2016-12-14
ScrollHandler can manage the center point of overlay components.

krris:
This PR adds a possibility to customise the center point of overlay components in `HUBViewControllerScrollHandler`.

### DIFF
--- a/include/HubFramework/HUBViewControllerScrollHandler.h
+++ b/include/HubFramework/HUBViewControllerScrollHandler.h
@@ -78,6 +78,17 @@
 - (UIEdgeInsets)contentInsetsForViewController:(HUBViewController *)viewController
                          proposedContentInsets:(UIEdgeInsets)proposedContentInsets;
 
+
+/**
+ *  Return the center point of an overlay coponent used in a view controller
+ *
+ *  @param viewController The view controller in question
+ *  @param proposedCenterPoint The center point that the Hub Framework is proposing
+ *
+ */
+- (CGPoint)centerPointForOverlayComponentInViewController:(HUBViewController *)viewController
+                                      proposedCenterPoint:(CGPoint)proposedCenterPoint;
+
 /**
  *  React to that a scrolling event started in a view controller
  *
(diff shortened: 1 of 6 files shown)


include/HubFramework/HUBViewControllerScrollHandler.h
```diff
@@ -78,6 +78,17 @@
...
 
+
+/**
+ *  Return the center point of an overlay coponent used in a view controller
+ *
+ *  @param viewController The view controller in question
+ *  @param proposedCenterPoint The center point that the Hub Framework is proposing
+ *
+ */
+- (CGPoint)centerPointForOverlayComponentInViewController:(HUBViewController *)viewController
```
Lumenela: @JohnSundell What do you think about it?

krris: @Lumenela So what would you recommend to do?

JohnSundell: I totally agree with @Lumenela, I don't see how this has anything to do with scrolling. We want to stick to the single responsibility principle as much as possible.

JohnSundell: @krris What problem is it that you want to solve?

krris: @JohnSundell I want to center an overlay component when the keyboard is visible. Probably this approach is not the best.

 I've just looked at the code once more again. I'm not sure if the following function in `HubViewController` works just like it should:
```
- (CGPoint)overlayComponentCenterPoint
{
    CGRect frame = self.view.bounds;
    frame.origin.y = self.collectionView.contentInset.top;
    frame.size.height -= self.visibleKeyboardHeight + CGRectGetMinY(frame);

    CGPoint proposedCenterPoint = CGPointMake(CGRectGetMidX(frame), CGRectGetMidY(frame));
    return [self.scrollHandler centerPointForOverlayComponentInViewController:self
                                                          proposedCenterPoint:proposedCenterPoint];
}
```
It doesn't compute correctly middle point if the keyboard is visible. Removing following lines would sort this problem because view automatically changes its size if the keyboard is displayed. 
```
    frame.origin.y = self.collectionView.contentInset.top;
    frame.size.height -= self.visibleKeyboardHeight + CGRectGetMinY(frame);
```

krris:
I'm closing this PR. We have some problems in our integration layer and we don't need to modify anything in HubFramework. Everything here works fine 🚀

PR #239, opened 2016-12-14
View controller: Don’t force image loading to be async

JohnSundell:
Previously, all image loading (even from cache) was forced to be async. The reason behind this was that we were using collection view cells to track what images belong to what component. But since a while back we’re now using component wrappers for that instead, which means that we can now allow some image loading to be synchronous (that’s up to `HUBImageLoader`).

This heavily reduces image flickering in view that are rapidly reloaded.


JohnSundell:
@spotify/objc-dev

PR #243, opened 2016-12-15
Customizable center point of overlay component

krris:
`HubViewControllerDelegate` can manage the center point of overlay components.

### DIFF
--- a/sources/HUBViewController.m
+++ b/sources/HUBViewController.m
@@ -1058,8 +1058,14 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState
     frame.size.height -= self.visibleKeyboardHeight + CGRectGetMinY(frame);
 
     CGPoint proposedCenterPoint = CGPointMake(CGRectGetMidX(frame), CGRectGetMidY(frame));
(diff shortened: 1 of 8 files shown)


krris:
Ping 🎾  @JohnSundell @nataliq

sources/HUBViewController.m
```diff
@@ -1051,7 +1051,10 @@ - (CGPoint)overlayComponentCenterPoint
     CGRect frame = self.view.bounds;
     frame.origin.y = self.collectionView.contentInset.top;
     frame.size.height -= self.visibleKeyboardHeight + CGRectGetMinY(frame);
-    return CGPointMake(CGRectGetMidX(frame), CGRectGetMidY(frame));
+
+    CGPoint proposedCenterPoint = CGPointMake(CGRectGetMidX(frame), CGRectGetMidY(frame));
+    return [self.delegate centerPointForOverlayComponentInViewController:self
```
JohnSundell: If the delegate is `nil` this will return `CGPointZero`, which is not what we want :)

krris: Right. I'll fix that.

krris:
@JohnSundell 🎾

PR #246, opened 2016-12-19
Make HUBAddJSONDataToBuilder() safer and rework error creation

rastersize:
This PR makes the helper `HUBAddJSONDataToBuilder()` function safer and reworks how we create errors.

### Make `HUBAddJSONDataToBuilder()` safer
This makes adding JSON data to a `HUBJSONCompatibleBuilder` safer. Since we’ll make sure `nil` isn’t passed to an Apple API that doesn’t support it.

Basically the Objective-C compiler is so easy to trick about nullability we’ve observed places where `nil` was passed to `addJSONData:error:`. This change makes sure we catch that before sending nil to an Apple owned API parameter that is marked as nonnull. Better safe than sorry eh 😅

### Rework how we create errors
This changes all errors to use a domain and code.

The standard Cocoa way of creating errors is to create an error domain that groups errors belonging together. The domain should not be used to identify _specific_ errors, as was done before this PR. Instead the combination of error domain and code should be used to identify a specific error. Which should then obviously be handled accordingly 😉 

I’ve translated the existing potential errors from the unique error domains they were using to domain + codes tuples.

---

Review would be appreciated @spotify/objc-dev 🛩

### DIFF
--- a/include/HubFramework/HUBErrors.h
+++ b/include/HubFramework/HUBErrors.h
@@ -1,0 +1,61 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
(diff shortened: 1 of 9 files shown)


include/HubFramework/HUBErrors.h
```diff
@@ -0,0 +1,63 @@
...
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+#pragma mark - JSON Serialization Errors
+
+#pragma mark Error Domain
+/// Error domain for JSON serialization errors.
+static NSErrorDomain const HUBJSONSerializationErrorDomain = @"com.spotify.hubframework.json-serialization";
```
8W9aG: Shouldn't this be an extern const?

rastersize: I was thinking that too, but then all string constants were static. So followed the same pattern. That said I think it’d make sense to change to extern const. Can do so here.

rastersize:
@8W9aG Updated to move the error domain constants to an implementation file making the declarations `extern`.

PR #248, opened 2016-12-20
README: Add link to Hub Framework Sessions Episode 1

JohnSundell:
More links will follow as we publish more videos


JohnSundell:
@spotify/objc-dev

PR #249, opened 2016-12-20
Remove code that is reseting the content offset when insets change

nataliq:
That code was probably a workaround for some issue but it is not covered by any tests and there are no comments. Given the fact that it has side effects causing bugs (if the collection view is scrolled to a specific position when the content insets change, the content offset will be reset) it is better if this code is removed.

### DIFF
--- a/sources/HUBViewController.m
+++ b/sources/HUBViewController.m
@@ -1141,14 +1141,8 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState
     contentInsets = [self.scrollHandler contentInsetsForViewController:self
                                                  proposedContentInsets:contentInsets];
 
-    if (!UIEdgeInsetsEqualToEdgeInsets(self.collectionView.contentInset, contentInsets)) {
-        self.collectionView.contentInset = contentInsets;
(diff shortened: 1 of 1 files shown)


Lumenela:
👍

rastersize:
@nataliq You seem to have added the code in f0ace389785b1b62609be3c4a76a129b1b3eaba5. The commit says:
> **Force update the content offset when content inset is set**
> I’ve found that sometimes when the content inset is set the content
> offset of the collection view is not updated for example when the view
> is not on the screen and is inside a child view controller..

Could you make sure that this problem isn’t re-introduced with the PR?

Lumenela:
@rastersize We've verified with @nataliq that the problem is not re-introduced in the PR.

Goos:
I'm not entirely familiar with the original issue, but wouldn't a more robust solution be to increase/decrease the content offset with the change in content inset? E.g. if the top content inset changes from `0` to `10`, decrease the current content offset with `10`. Or does the `UICollectionView` take care of that by itself?

Lumenela:
Collection view manages content offset when content inset changes

PR #251, opened 2016-12-21
View controller: Make rendering code operation oriented

JohnSundell:
This is the first in a series of tasks to make more code in the Hub Framework operation oriented.

We’ve run into a lot of issues during the last couple of months due to state management, where things have become out of sync or required even more state to be introduced. For example, when interfacing with `UICollectionView`, which is very asynchronous, we could end up with inconsistent states because we would update our models too early or too late.

Operations make this a lot simpler, since using a central queue that all events pass through, you can make more assumptions (and IMO, reason about it easier) about in what sequence various events and bodies of work will occur.

Now, instead of having to keep state like `pendingViewModel` and `isRendering`, `HUBViewController` now has an operation queue that it adds operations onto whenever it needs to re-render.

To make this happen, two new classes are introduced: `HUBOperation` and `HUBOperationQueue`. `NSOperations` where considered, but ruled out due to their complexity and optimisation towards background thread work (see the documentation for `HUBOperation` for a more in-depth discussion). So why are these not named `HUBRenderingOperation` (+queue)? Because we want to migrate more code to using this paradigm, but exclusively rendering code. For example, once we move to context-based content operations (see `HUBContentOperationContext`), we could implement a simple input/output for `HUBOperation` and use the same code for those.

But, for now, only the rendering code is updated.

### DIFF
--- a/sources/HUBOperationQueue.m
+++ b/sources/HUBOperationQueue.m
@@ -1,0 +1,92 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import "HUBOperationQueue.h"
+
+#import "HUBOperation.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface HUBOperationQueue ()
+
+@property (nonatomic, strong, readonly) NSMutableArray<HUBOperation *> *operations;
+
+@end
+
+@implementation HUBOperationQueue
+
+#pragma mark - Initializer
+
+- (instancetype)init
+{
+    self = [super init];
+    
+    if (self) {
+        _operations = [NSMutableArray new];
+    }
+    
+    return self;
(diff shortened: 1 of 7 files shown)


JohnSundell:
@spotify/objc-dev

sources/HUBOperationQueue.m
```diff
@@ -0,0 +1,92 @@
...
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import "HUBOperationQueue.h"
+
```
ynmehel: Is the line between import statements intentional here? Can be removed

sources/HUBOperationQueue.m
```diff
@@ -0,0 +1,92 @@
...
+    __weak __typeof(self) weakSelf = self;
+    
+    [operation performWithCompletionHandler:^{
+        __typeof(self) strongSelf = weakSelf;
+        
+        if (strongSelf == nil) {
+            return;
+        }
+        
+        [strongSelf.operations removeObjectAtIndex:0];
```
Goos: Perhaps we should use a more thread-safe way of manipulating the operation queue here, as we could technically insert an operation on the main queue while popping one in the completion handler (which, I assume, is up to the asynchronous block to call on whichever thread it pleases?).

JohnSundell: `HUBOperationQueue` is designed to be single theaded for now though. But could wrap it in `HUBPerformOnMainQueue` to guard us if any background operations are introduced in the future.

sources/HUBOperationQueue.m
JohnSundell: Yeah, it's intentional. The first is the header matching the implementation file, the second is the first dependency 🙂

JohnSundell:
@Goos Would rather not move around the code too much, to keep the impact small. It would be great to move all of the rendering-related code into `HUBViewModelRenderer` (it could keep its own operation queue even) in a subsequent PR though.

sources/HUBOperationQueue.m
JohnSundell: Added this: https://github.com/spotify/HubFramework/pull/251/commits/94208ee3990fae99007d14bd2d344a5a8d65f27a 😄

Goos: 👌 👌

Issue #252, opened 2016-12-21
Unify parameters passed to content operations


rastersize:
In the PR “Introduce HUBBlockContentOperation + context” (#199) @JohnSundell introduced the `HUBContentOperationContext` protocol which encapsulates all values passed to a content operation. However the PR only made it available to the block based content operation. The intention is to pass an object of this type to all content operations. Instead of sending the values as individual parameters.

The main reason we want to move to a context object over individual parameters is due to scalability. Continuing to add parameters to the `performWith…` method will quickly make it hard to work with and maintain. We also want to keep the way API consumers write content operations the coherent between “normal” (custom class) operation and block based operations.

Furthermore the `HUBContentOperationWithPaginatedContent` protocol declares an almost identical, but still different, method. Which requires us to keep its `append` method parameters in sync.

## Proposal
I propose that we refactor the `perform` method of `HUBContentOperation` and the `append` method of `HUBContentOperationWithPaginatedContent` to using a context parameter. Instead of their existing parameters. This change would be an API breaking change. However it will insulate us such that we can make future changes to the API of what a content operation is provided.

API wise the change would be like follows.

### `HUBContentOperation`
```objc
@protocol HUBContentOperation <NSObject>
// Before
- (void)performForViewURI:(NSURL *)viewURI
              featureInfo:(id<HUBFeatureInfo>)featureInfo
        connectivityState:(HUBConnectivityState)connectivityState
         viewModelBuilder:(id<HUBViewModelBuilder>)viewModelBuilder
            previousError:(nullable NSError *)previousError;

// After
- (void)performInContext:(id<HUBContentOperationContext>)context;
@end
```
```swift
// Imports in Swift as
func perform(in context: HUBContentOperationContext)
```

The consumer side the change results in that existing code will need to prepend any use of `viewURI`, 
`featureInfo`, `connectivityState`, `viewModelBuilder` and `previousError` with `context.`. Obviously after changing the method definition 😉

### `HUBContentOperationWithPaginatedContent`
```objc
@protocol HUBContentOperationWithPaginatedContent <NSObject>
// Before
- (void)appendContentForPageIndex:(NSUInteger)pageIndex
               toViewModelBuilder:(id<HUBViewModelBuilder>)viewModelBuilder
                          viewURI:(NSURL *)viewURI
                      featureInfo:(id<HUBFeatureInfo>)featureInfo
                connectivityState:(HUBConnectivityState)connectivityState
[...]

PR #253, opened 2016-12-21
Content operation context everywhere

rastersize:
This PR is a concrete implementation of #252. Unifying how parameters are passed to content operations.

With this PR all content operations will be given the same type of parameter. Encapsulating the entire context in which the operation runs. Yay unity!

To do this we, unfortunately, have to break the existing API of `HUBContentOperation`. However this PR makes the framework easier to change in the future. So that we don’t need to break the API whenever a new context parameter needs to be introduced or an existing one changed.

The changes needed for consumers of the API are quite small and simple. As is visible from the changes to the demo project. The changes are more or less just updating to the new method name and then prefixing the parameters with `context.` 🎉 

---

The PR also adds tests making sure `HUBBlockContentOperation` actually sends the `contentOperationDidFinish:` message to the delegate 😸


sources/HUBContentOperationWrapper.m
```diff
@@ -56,29 +58,30 @@ - (void)performOperationForViewURI:(NSURL *)viewURI
                      previousError:(nullable NSError *)previousError
 {
     self.isExecuting = YES;
-    
+
+    id<HUBContentOperationContext> const context = ({
```
Goos: What's the point of the `({ })`-scoping here? Haven't seen that in obj-c before 😮

rastersize: It’s a GNU C thingy (“[statement expression](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html)”) that basically creates a new scope and then the last value is “returned”. The use-case here is to “trick” Xcode into indenting the statement in a way that’s nice to read. Which is all good since we use gnu11 as our C dialect :)

Goos: Aha, cool :smile:

the more you know

PR #255, opened 2016-12-21
Update Fastlane to 2.4.0 from 1.111

rastersize:
They’ve moved everything into a single gem and also made a lot of releases over the path month. So we’re att ~~2.2.0~~ 2.4.0 now :)

Since they’ve results in our `Gemfile.lock` being vastly shorter 🎉

@spotify/objc-dev


rastersize:
This is blocked on https://github.com/KrauseFx/trainer needing to push v0.4.0 to ruby gems. Version 0.3.0 depends on `fastlane_core` (i.e. pre-mono gem) so isn’t compatible with fastlane 2.

edvanbeinum:
👍

PR #256, opened 2016-12-22
Add test for content operation and render overlap

cerihughes:
Add a test that makes sure content operations and layout requests don't make overlapping calls to the render logic.

### DIFF
--- a/tests/HUBViewControllerTests.m
+++ b/tests/HUBViewControllerTests.m
@@ -3101,7 +3101,7 @@
     viewModelRenderer.completionBlocks[1]();
 }
 
-- (void)testThatModelIsNotSetImmediatelyOnOverlappingRenderAndLayoutRequests2
+- (void)testThatModelIsNotSetImmediatelyOnOverlappingRenderAndLayoutRequests
 {
     HUBViewModelRendererMock *viewModelRenderer = [HUBViewModelRendererMock new];
     [self createViewControllerWithViewModelRenderer:viewModelRenderer];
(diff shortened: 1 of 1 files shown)


cerihughes:
@JohnSundell

tests/HUBViewControllerTests.m
```diff
@@ -3101,6 +3101,58 @@ - (void)testThatModelIsNotSetImmediatelyOnOverlappingRenderRequests
...
+        return YES;
+    };
+
+    // While the 1st render is in progress, a 2nd render request comes in
+    [self.viewController reload];
+
+    // The 2nd model should NOT be set at this point as the 1st render hasn't finished
+    XCTAssertEqualObjects(self.viewController.viewModel.headerComponentModel.title, @"title1");
+    // Only 1 render request has been made at this point (the 2nd is pending)
+    XCTAssertEqual(viewModelRenderer.completionBlocks.count, (NSUInteger)1);
```
JohnSundell: You can use `1u` to avoid the casting (same for the other integer literals).

tests/HUBViewControllerTests.m
```diff
@@ -3101,6 +3101,58 @@ - (void)testThatModelIsNotSetImmediatelyOnOverlappingRenderRequests
...
+    // The 2nd model should now be set as the 1st render has finished and the 2nd is in progress
+    XCTAssertEqualObjects(self.viewController.viewModel.headerComponentModel.title, @"title2");
+    // 2 render requests have now been made
+    XCTAssertEqual(viewModelRenderer.completionBlocks.count, (NSUInteger)2);
+
+    // While the 2nd render is in progress, a layout requests comes in.
+    CGRect rect = self.viewController.view.bounds;
+    CGRect offsetRect = CGRectOffset(rect, 1, 1);
+    self.viewController.view.bounds = offsetRect;
+    [self.viewController.view layoutIfNeeded];
```
JohnSundell: Wouldn't it be better to call `[viewController layoutSubviews]` directly? That way you're sure of the execution flow, so the test becomes more predictable.

JohnSundell:
Could the commit history be cleaned up a bit? Maybe squash all of them? 😇

tests/HUBViewControllerTests.m
cerihughes: Ooops. Cut and pasted that. Will fix, and also create a separate PR to clean this up everywhere else.

tests/HUBViewControllerTests.m
cerihughes: Will do.

cerihughes: Oh, actually, we can't do that as we need to trigger `viewDidLayoutSubviews` which isn't called when you directly call `layoutSubviews`.

JohnSundell: Sure, sounds good 👍

cerihughes:
Screwed up my branch trying to squash 😱 
Replacing this with https://github.com/spotify/HubFramework/pull/270

PR #259, opened 2016-12-22
Use more appropriate asserts

cerihughes:
Always use `XCTAssertEqualObjects` when comparing objects for equality. `XCTAssertEqual` was causing some issues with `NSIndexPath` equality on 32 bit architectures (possibly because 64-bit architectures pool instances that are equal?)

### DIFF
--- a/tests/HUBActionRegistryTests.m
+++ b/tests/HUBActionRegistryTests.m
@@ -58,7 +58,7 @@
     HUBActionFactoryMock * const factory = [[HUBActionFactoryMock alloc] initWithActions:@{actionIdentifier.namePart: action}];
     [self.actionRegistry registerActionFactory:factory forNamespace:actionIdentifier.namespacePart];
     
-    XCTAssertEqual([self.actionRegistry createCustomActionForIdentifier:actionIdentifier], action);
+    XCTAssertEqualObjects([self.actionRegistry createCustomActionForIdentifier:actionIdentifier], action);
 }
 
 - (void)testRegisteringAlreadyRegisteredFactoryThrows
--- a/tests/HUBViewModelRendererTests.m
+++ b/tests/HUBViewModelRendererTests.m
@@ -108,8 +108,8 @@
     [self waitForExpectationsWithTimeout:2 handler:nil];
 
     XCTAssertEqual([self.collectionViewLayout numberOfInvocations], 2u);
-    XCTAssertEqualObjects([self.collectionViewLayout capturedViewModelAtIndex:0], firstViewModel);
-    XCTAssertEqualObjects([self.collectionViewLayout capturedViewModelAtIndex:1], secondViewModel);
+    XCTAssertEqual([self.collectionViewLayout capturedViewModelAtIndex:0], firstViewModel);
+    XCTAssertEqual([self.collectionViewLayout capturedViewModelAtIndex:1], secondViewModel);
 
     XCTAssertEqual([self.collectionViewLayout numberOfInvocations], 2u);
     // The first invocation shouldn't generate a diff.
--- a/tests/HUBComponentGestureRecognizerTests.m
+++ b/tests/HUBComponentGestureRecognizerTests.m
@@ -57,7 +57,7 @@
 
 - (void)testGestureRecognizerAddedToView
 {
-    XCTAssertEqual(self.gestureRecognizer.view, self.view);
+    XCTAssertEqualObjects(self.gestureRecognizer.view, self.view);
 }
 
 - (void)testTouchesBeganSetsBeganState
--- a/tests/HUBComponentImageDataBuilderTests.m
+++ b/tests/HUBComponentImageDataBuilderTests.m
@@ -74,10 +74,10 @@
     
     HUBComponentImageDataImplementation * const imageData = [self.builder buildWithIdentifier:identifier type:type];
     
-    XCTAssertEqual(imageData.identifier, identifier);
+    XCTAssertEqualObjects(imageData.identifier, identifier);
     XCTAssertEqual(imageData.type, type);
     XCTAssertEqualObjects(imageData.URL, self.builder.URL);
-    XCTAssertEqual(imageData.localImage, self.builder.localImage);
+    XCTAssertEqualObjects(imageData.localImage, self.builder.localImage);
     XCTAssertEqualObjects(imageData.placeholderIcon.identifier, @"placeholder");
     XCTAssertEqualObjects(imageData.customData, @{@"custom": @"data"});
 }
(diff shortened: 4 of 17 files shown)


tests/HUBActionRegistryTests.m
```diff
@@ -58,7 +58,7 @@ - (void)testRegisteringFactoryAndCreatingAction
     HUBActionFactoryMock * const factory = [[HUBActionFactoryMock alloc] initWithActions:@{actionIdentifier.namePart: action}];
     [self.actionRegistry registerActionFactory:factory forNamespace:actionIdentifier.namespacePart];
     
-    XCTAssertEqual([self.actionRegistry createCustomActionForIdentifier:actionIdentifier], action);
+    XCTAssertEqualObjects([self.actionRegistry createCustomActionForIdentifier:actionIdentifier], action);
```
JohnSundell: Here we actually want to make sure that the objects are the same **instance** so `XCTAssertEqual` should be used.

cerihughes:
@spotify/objc-dev

tests/HUBActionRegistryTests.m
cerihughes: I wonder if we should use `XCTAssertTrue([self.actionRegistry createCustomActionForIdentifier:actionIdentifier] == action)` for that then, just to be explicit about it.

cerihughes: Hamcrest has a "same instance" matcher that is good for these scenarios. XCTest doesn't seem to though :(

JohnSundell: Isn't it more explicit to use `XCTAssertEqual` that does exactly that? 😅

cerihughes: Yeah, possibly. I have a problem with `XCTAssertEqual` as it's often misused, and people often aren't asserting what they think. If you *do* want to test instance equality, then using `XCTAssertEqual` just feels wrong as actually tests "more" than equality.

cerihughes: If we are testing instance equality `XCTAssertTrue(a == b)` is far more explicit.

tests/HUBActionRegistryTests.m
```diff
@@ -57,8 +57,9 @@ - (void)testRegisteringFactoryAndCreatingAction
     HUBActionMock * const action = [[HUBActionMock alloc] initWithBlock:nil];
     HUBActionFactoryMock * const factory = [[HUBActionFactoryMock alloc] initWithActions:@{actionIdentifier.namePart: action}];
     [self.actionRegistry registerActionFactory:factory forNamespace:actionIdentifier.namespacePart];
-    
-    XCTAssertEqual([self.actionRegistry createCustomActionForIdentifier:actionIdentifier], action);
+
+    // These should be the same instance.
+    XCTAssertTrue([self.actionRegistry createCustomActionForIdentifier:actionIdentifier] == action);
```
JohnSundell: I really fail to understand why this is better than `XCTAssertEqual` which is meant for exactly these type of comparisons.

JensAyton: I agree with Ceri; `XCTAssertEqual` is a smell for non-scalar types.

cerihughes: I guess it's because in a test, if I look at a line that says `XCTAssertEqual(objA, obj2)`, I immediately think "do we really want to check equality here, or is this a typo".

Admittedly `XCTAssertTrue(obj1 == obj2)` doesn't do a great deal to fix that ambiguity, which is why I like Hamcrest's approach of providing a "same instance" matcher for objects so that it's obvious by looking at it whether you want to check for equality or instance equivalence.

cerihughes: I think the comment is the most useful thing here. I'll revert the assert and keep the comment.

cerihughes:
@JohnSundell : I would assume that by default we'd want to be checking object equality in tests, unless we have a requirement that something returns the same instance under certain conditions. Only then would we want to test the actual instances.

The `NSIndexPath` case is a good example - on some platforms (for some reason) calls to factory methods will return different instances, and on other platforms they return the same instance (I haven't verified this, but it's one explanation for what's happening). Ultimately though, it shouldn't matter, as long as we use object equality, not instance equality as `isEqualTo:` will take care of all the differences for us.

Only fixing the cases with `NSIndexPath` because we know there are problems with that right now seems a bit broken, as the same issue could arise with other classes in future OS releases (as an example).

JensAyton:
@cerihughes `NSIndexPath`s of two elements are tagged pointers on 64-bit systems. Same sort of fun happens with short strings.

tests/HUBActionRegistryTests.m
rastersize: How about adding a description to the assert? That is:
```objc
XCTAssertTrue([self.actionRegistry createCustomActionForIdentifier:actionIdentifier] == action, @"The action should be the same instance as the factory mock vends");
```

That way the comment will be directly attached to the assert. Making sure it also shows up when the test fails.

tests/HUBViewModelRendererTests.m
```diff
@@ -108,8 +108,8 @@ - (void)testTwoSubsequentRenders
     [self waitForExpectationsWithTimeout:2 handler:nil];
 
     XCTAssertEqual([self.collectionViewLayout numberOfInvocations], 2u);
-    XCTAssertEqualObjects([self.collectionViewLayout capturedViewModelAtIndex:0], firstViewModel);
-    XCTAssertEqualObjects([self.collectionViewLayout capturedViewModelAtIndex:1], secondViewModel);
+    XCTAssertEqual([self.collectionViewLayout capturedViewModelAtIndex:0], firstViewModel);
```
rastersize: Shouldn’t we use `XCTAssertTrue([self.collectionViewLayout capturedViewModelAtIndex:0] == firstViewModel)` here as well? As well as adding an assert description stating that we really do want to check for instance equality.

Just like at https://github.com/spotify/HubFramework/pull/259/files#diff-51a93e5f57aabdf96df5027e93c75d58R62

tests/HUBViewModelRendererTests.m
```diff
@@ -108,8 +108,8 @@ - (void)testTwoSubsequentRenders
...
+    XCTAssertEqual([self.collectionViewLayout capturedViewModelAtIndex:1], secondViewModel);
```
rastersize: Same as https://github.com/spotify/HubFramework/pull/259/files#r94663198

rastersize:
@cerihughes I’m in favor of the changes made in this PR. Unless we explicitly want to check that two objects are the same instance we should use `XCTAssertEqualObjects(obj1, obj2, desc, …)`.

Further I think we should, like you’re doing here, use `XCTAssertTrue()` with a description (or comment) when we actually want to check for instance equality. I’d also be fine with adding a function or macro like `HUBAssertEqualInstances(ins1, ins2, desc, …)` to make the code more self-explanatory.

Issue #164, opened 2016-11-16
Support tvOS


JohnSundell:
A commonly requested feature is for Hub Framework to support tvOS. Since the framework just uses standard UIKit classes, it shouldn't be too hard (famous last words 😅). It should involve the following steps:

- [ ] Enable the Hub Framework to be compiled for a tvOS target.
- [ ] Add support for the focus engine.
- [ ] Update the CI to build for tvOS as well as iOS.

JensAyton:
The Hub Framework is being phased out at Spotify, and therefore we will not be maintaining it further.

PR #263, opened 2016-12-26
Add tvOS support

marmelroy:
Fixes https://github.com/spotify/HubFramework/issues/164.

There are several things going on here:
- Made HubFramework compile for tvOS (lots of compile flags).
- Added support for the focus engine (a little trickier than expected, especially with nested collection views).
- Added a HubComponentWithFocusState protocol so that a component can update its view when in focus. 
- Added a tvOS test target. 

Missing:
- I did not update the CI setup. I'm not super familiar with using fastlane for this (often use a simple shell script) and wanted to avoid breaking something. 
- A tvOS demo app (to come in a separate PR later on).


tests/HUBViewControllerTests.m
```diff
@@ -1849,6 +1849,7 @@ - (void)testRenderingUpdatesContentInsetBeforeAndAfterRendering
 
 - (void)testProposedContentInsetIsDefaultIfHeaderMissing
 {
+    #if !TARGET_OS_TV
```
cerihughes: Should this encompass the whole test (including the method signature) as with `testAdaptingOverlayComponentCenterPointToKeyboard`?

marmelroy:
Thanks for the feedback.

A universal framework is much easier to maintain and that's a huge plus. The main technical argument against a universal framework was that Carthage couldn't build it but that's no longer the case. 

See these for context:
https://github.com/Alamofire/Alamofire/pull/326
https://github.com/Alamofire/Alamofire/pull/1064

Any other argument we should consider in favour of a platform-specific framework? 

An additional difference with @JohnSundell's implementation is that I added an explicit HubComponentWithFocusState protocol whereas John merged focus state with highlighted state. 

I will update the PR with the changes requested by @rastersize and @cerihughes and will steal a couple of ideas from @JohnSundell's branch 😄

rastersize:
@marmelroy Thank you for the links and update, much appreciated! From reading the Alamofire discussions I think we should probably go with multiple targets. Like you’ve done in this PR.

include/HubFramework/HUBComponentWithFocusState.h
```diff
@@ -0,0 +1,50 @@
...
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+#import "HUBComponent.h"
+
+/// Enum defining various focus states that a component can be in
+typedef NS_ENUM(NSUInteger, HUBComponentFocusState) {
```
marmelroy: Erm... I was just being consistent with most of the other NS_ENUMs in the project (see HUBComponentSelectionState, HUBActionTrigger, HUBComponentLayoutContentEdge, HUBComponentType, HUBConnectivityState etc). 

There's no specific reason I can think of for any of them to be unsigned. Should we change all enums?

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -2128,13 +2800,172 @@
 /* End PBXTargetDependency section */
 
 /* Begin XCBuildConfiguration section */
+		344C5F101E09CF9A00597B61 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				ALWAYS_SEARCH_USER_PATHS = NO;
```
marmelroy: Finally had the chance to look into this. So... I basically just duplicated the setup of the current HubFramework-iOS dynamic target and made my tvOS modifications. All of these overrides are currently in production. 

I will make this modification to both dynamic targets then.

marmelroy:
Sorry for the delay in fixing this - moving countries is quite time exhaustive. This PR should be ready for another review. 

Summary of changes:
- Syntax was improved as per @cerihughes request. 
- Change requested re unsigned enums was applied framework-wide. 
- Change requested re configuration overrides was due to duplication of the dynamic target on production. Overrides were removed from both iOS and tvOS targets.
- The test target was improved a little.

mhallendal:
I would prefer we wait with this until we have the API more stable, we also need to refactor `HUBViewController` to do less and be easier to reason about which will become more difficult by having to care for tvOS as well.

PR #265, opened 2016-12-30
Remove NSObject inheritance from all protocols.

marmelroy:
I was intrigued by the discussion on https://github.com/spotify/HubFramework/issues/70.

This PR solves the main issue with removing NSObject inheritance (by providing an NSObject-agnostic HUBConformsToProtocol function in HUBUtilities). The only protocol that keeps NSObject is HUBSerializable because it needs isEqual. 

Quick note: The reason HUBConformsToProtocol first tries to use the object's confromsToProtocol implementation (if it exists) is to make sure we don't break mock classes that override it. 

Missing:
- Linting.

### DIFF
--- a/sources/HUBUtilities.h
+++ b/sources/HUBUtilities.h
@@ -20,6 +20,7 @@
  */
 
 #import <UIKit/UIKit.h>
+#import <objc/runtime.h>
 
 #import "HUBComponent.h"
 #import "HUBErrors.h"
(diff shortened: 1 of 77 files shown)


rastersize:
Had a quick look and looks good, I’ll give it a proper review on Tuesday when I’m back in the office. Thank you for the PR much appreciated!

rastersize:
@marmelroy We can add the linting later. Since it’ll require some infrastructure to be added 😄

marmelroy:
Thanks @rastersize. Good catch.

cerihughes:
I don't see any tests for `HUBConformsToProtocol`. Can we add some, considering it's now an important part of our codebase?

sources/HUBUtilities.h
```diff
@@ -51,6 +52,23 @@ static inline BOOL HUBPropertyIsEqual(NSObject * _Nullable objectA, NSObject * _
...
 
 /**
+ *  Return whether an object conforms to a protocol
+ *
+ *  @param object The object
+ *  @param aProtocol The protocol
+ *
+ *  This function allows checking protocol conformance without inheriting from NSObject.
+ */
+static inline BOOL HUBConformsToProtocol(id _Nullable object, Protocol *aProtocol)
```
cerihughes: I don't see any tests for `HUBConformsToProtocol`. Can we add some, considering it's now an important part of our codebase?

marmelroy:
@cerihughes, just got around to adding a simple test that covers all the cases. Let me know your thoughts.

marmelroy:
🎉  shipit

rastersize:
Niiice 😃 🌮 🎉

PR #268, opened 2017-01-03
Don’t use XCTAssertEqual for NSIndexPath equivalence.

cerihughes:
While we decide the best way to assert for object equivalence over at https://github.com/spotify/HubFramework/pull/259, we should fix the cases that are causing tests to fail on 32-bit systems (i.e. the `NSIndexPath` case).

For info, using instance equality (i.e. `==` ) on 2 "equivalent" instances will work on 64 bit architectures (with tagged pointers) but not on 32 bit architectures.

### DIFF
--- a/tests/HUBComponentModelTests.m
+++ b/tests/HUBComponentModelTests.m
@@ -170,16 +170,16 @@
     parent.children = @[childA, childB];
     childB.children = @[grandchild];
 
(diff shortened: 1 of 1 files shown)


cerihughes:
@spotify/objc-dev

dalef84:
👍

PR #277, opened 2017-01-23
Don’t reload the collection view if there is nothing in the diff.

cerihughes:
When we calculate a diff and there are no changes, we don't need to spend time reloading the collection view, either with `reloadData` or `performBatchUpdates:completion:`. These methods are already called too often and have been linked to crashes under certain conditions where we update content operations frequently. After I applied this patch, I didn't see any crashes under those same conditions.

The `HUBViewModelRenderer` patch is quite hard to follow, so here's a summary of what I did:

- Created a new `renderViewModel:inCollectionView:usingBatchUpdates:addHeaderMargin:completion:` method (i.e. the same as the existing method but without the `animated:` parameter.
- In the original method (with `animated:`, we now call the new method in a block, either directly (if we are animating), or through `[UIView performWithoutAnimation:renderBlock]` if we're not. This handles all animation specifics, allowing the new method to not care about that part.
- Moved all remaining logic to the new method.
- There are 4 common operations that happen in this class:
  - reloading the collection view
  - updating the layout
  - updating the view model
  - firing the completion
- Due to the nature of the calls to the collection view, the last 3 operations are called at different times, but we can break them down as
  - layout tasks (updating the layout)
  - post-layout tasks (updating the view model and firing the completion)
- The PR defines those 2 tasks as blocks and calls them from all of the same places where they used to be called from.
- The PR also now checks if the diff actually has any changes, and if it doesn't, it doesn't update the collection view - it just triggers the 2 blocks.

### DIFF
--- a/sources/HUBViewModelDiff.h
+++ b/sources/HUBViewModelDiff.h
@@ -70,6 +70,9 @@ extern HUBViewModelDiff *HUBDiffMyersAlgorithm(id<HUBViewModel>, id<HUBViewModel
 /// The index paths of any body components that were modified in the new view model. 
 @property (nonatomic, strong, readonly) NSArray<NSIndexPath *> *reloadedBodyComponentIndexPaths;
 
+/// A convenience property that returns YES if there are any inserts, deletes or reloads in this diff.
+@property (nonatomic, readonly) BOOL hasChanges;
+
 /**
  * Initializes a @c HUBViewModelDiff using the two view models by finding the longest common subsequence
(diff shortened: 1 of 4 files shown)


cerihughes:
@Goos @mhallendal

mhallendal:
Let's try it out and see how it works over a longer period of time. It sounds like this may reduce the risk but not solve the root cause but let's go with this until we have time to get to the bottom of the rendering logic.

👍

PR #278, opened 2017-02-02
Don’t register multiple times

cerihughes:
`self.registeredCollectionViewCellReuseIdentifiers` was being set up and checked for an existing cell reuse identifier, but we never added the identifier to the collection to stop us registering multiple times.


cerihughes:
@spotify/objc-dev

marmelroy:
👍

PR #279, opened 2017-02-13
Support components having different layouts on rotation.

cerihughes:
When the `HUBViewController` creates a cell, it creates the correct component for that cell and calls `configureViewWithModel:containerViewSize:` on that component, which can use this information to create a component-specific layout that takes the parent size into consideration.

However, when the VC is rotated / the VC is resized in other ways, there's no opportunity to recalculate that layout with the new parent size information.

This PR addresses this by forwarding this information to all component wrappers currently being displayed by the view controller's collection view.


cerihughes:
@spotify/objc-dev

PR #280, opened 2017-02-14
Don’t discard important objects in low memory conditions

cerihughes:
Setting the view model and collection view to `nil` under low memory conditions doesn't seem like a very safe thing to do. We're seeing crashes because:
-  collection views are reporting that they don't have a reuse identifier registered 
- collection views are reporting discrepancies between models and cells currently visible

Both could be caused by a collection view / view model being set to `nil` and recreated at some point in the middle of an important operation.

Let's try removing the `didReceiveMemoryWarning` implementation and see if we see a reduction in crash reports.

### DIFF
--- a/sources/HUBViewController.m
+++ b/sources/HUBViewController.m
@@ -250,29 +250,6 @@ NS_ASSUME_NONNULL_BEGIN
     } completion:nil];
 }
 
-- (void)didReceiveMemoryWarning
-{
-    [super didReceiveMemoryWarning];
-
-    if (!self.isViewLoaded) {
-        return;
-    }
-
-    if (self.view.window != nil) {
-        return;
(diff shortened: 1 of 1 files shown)


cerihughes:
@mhallendal @spotify/objc-dev

mhallendal:
Nice find, there is definitely state that will screw up in the ways you described by throwing away only the `collectionView` and `viewModel`. 

`registeredCollectionViewCellReuseIdentifiersregisteredCollectionViewCellReuseIdentifiers` keeps track of registered classes and will contain classes registered with a deallocated `UICollectionView` and the `HUBViewRenderer` keeps track of the last view model so it will calculate a diff against a different view model than the `HUBViewController` knows about.

PR #283, opened 2017-02-14
Enables UISegmentedControl touch

mumme:
@rastersize

### DIFF
--- a/sources/HUBComponentWrapper.m
+++ b/sources/HUBComponentWrapper.m
@@ -432,6 +432,11 @@ NS_ASSUME_NONNULL_BEGIN
     UIView *currentView = touch.view;
     
     while (currentView != nil && currentView != self.view) {
+
+        if ([currentView isKindOfClass:[UISegmentedControl class]]) {
+            return NO;
+        }
+
         if ([currentView isKindOfClass:[UIButton class]]) {
             return NO;
         }


sources/HUBComponentWrapper.m
```diff
@@ -432,6 +432,11 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceive
     UIView *currentView = touch.view;
     
     while (currentView != nil && currentView != self.view) {
+
```
rastersize: 🎨 Remove the empty line

sources/HUBComponentWrapper.m
```diff
@@ -432,6 +432,11 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceive
...
+        if ([currentView isKindOfClass:[UISegmentedControl class]]) {
```
rastersize: Wondering if we should not add this check and instead change the `UIButton` check to `UIControl`?

@mhallendal @cerihughes What do you and CA think?

marmelroy:
I was intrigued by the fact that we were traversing the view hierarchy looking for specific subclasses and created an alternative PR - https://github.com/spotify/HubFramework/pull/285

rastersize:
I’d be in favour of merging this (after changing to look for `UIControl`) and the concept of #285 into a single change. That is:
1. Look for a gesture controller ala. #285 if not found proceed to,
2. Go through the view hierarchy and look for a `UIControl`.

rastersize:
@mhallendal @marmelroy @mumme What do you think of this ☝️

PR #285, opened 2017-02-15
Alternative to https://github.com/spotify/HubFramework/pull/283

marmelroy:
Traversing the view hierarchy in search of specific class that handle touches is not really scalable. 

This is an alternative solution.


rastersize:
@mumme @marmelroy I prefer this solution a lot to what Hub framework does and what was added in #283! Seems like a better way to do it.

mumme:
This solution did not work for the segmented control case.

marmelroy:
@mumme, did a little investigation and it appears that the reason this doesn't work is that UISegementedControl doesn't implement any gesture recognisers and captures touches directly (thanks Apple).

PR #286, opened 2017-02-15
Tie cell class register to collection view.

mhallendal:
As noticed in #280 the registry that keeps track of registered reuse identifiers need to be better tied to the collection view. This makes it a property of `HUBCollectionView` which means that whenever we recreate the collection view in `HUBViewController` the registry will be recreated as well.

@spotify/objc-dev @cerihughes

### DIFF
--- a/sources/HUBCollectionView.m
+++ b/sources/HUBCollectionView.m
@@ -23,8 +23,12 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@implementation HUBCollectionView
+@interface HUBCollectionView ()
+@property (nonatomic, strong, readonly) NSMutableSet<NSString *> *registeredCollectionViewCellReuseIdentifiers;
+@end
 
+@implementation HUBCollectionView
+@synthesize registeredCollectionViewCellReuseIdentifiers = _registeredCollectionViewCellReuseIdentifiers;
 @dynamic delegate;
 
 - (void)setContentOffset:(CGPoint)contentOffset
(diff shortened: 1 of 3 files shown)


sources/HUBCollectionView.m
```diff
@@ -42,6 +46,27 @@ - (void)setContentOffset:(CGPoint)contentOffset
...
+- (NSMutableSet<NSString *> *)registeredCollectionViewCellReuseIdentifiers
+{
+    if (!_registeredCollectionViewCellReuseIdentifiers) {
+        _registeredCollectionViewCellReuseIdentifiers = [NSMutableSet new];
+    }
+
+    return _registeredCollectionViewCellReuseIdentifiers;
+}
+
+- (__kindof UICollectionViewCell *)dequeueReusableCellWithReuseIdentifier:(NSString *)identifier
```
cerihughes: I don't really have a strong opinion, but would it be better to just add this logic into an overridden `registerClass: forCellWithReuseIdentifier:`? That way we wouldn't have to add to the public API of this class.

sources/HUBCollectionView.m
```diff
@@ -42,6 +46,27 @@ - (void)setContentOffset:(CGPoint)contentOffset
...
+                                                             forIndexPath:(NSIndexPath *)indexPath
+                                                cellClassWhenUnregistered:(Class)cellClass
+{
+    if (![self.registeredCollectionViewCellReuseIdentifiers containsObject:identifier]) {
+        [self registerClass:cellClass forCellWithReuseIdentifier:identifier];
```
rastersize: How about doing this in the `loadCollectionViewIfNeeded` method? Would also allow us to not expose any new API. Or is there a specific reason we want to do it this way (i.e. JIT for when the cell is being dequeued)? Feels a bit scary and on the edge of the API constraints.

mhallendal: Not sure I follow here, we don't know all the potential identifiers in `loadCollectionViewIfNeeded`. This code is just copied over from `collectionView:cellForItemAtIndexPath:` which is where it's being called from still.

sources/HUBCollectionView.m
mhallendal: I didn't want to change the semantics of that method in the subclass and since it's an internal class anyway I figured it's better to make it behave the same way and add another call for when we want specific behaviour.

PR #288, opened 2017-02-20
Fix building the static library with Xcode 8.3 beta 3

rastersize:
Fix building with Xcode 8.3 and by doing so fix all 32 vs 64 bit `CGFloat` issues. The new Xcode version once again bumped into our “ignore warnings per Xcode version” hack. Instead of applying the easy fix I opted to instead fix all 32 vs 64 bit issues.

### Remove all usage of the `f` floating point suffix
The usage of the `f` suffix results in a wonkiness with `float` values either being implicitly converted to `double`s. Or in a worse scenario causing `double` values to lose precision implicitly in certain operations.

In general we should never use the floating point specifier (i.e. the `f` suffix). Unless it’s an edge case where it’s actually needed. We should let the compiler figure out which trivial type should be used.

### Fix 32 vs. 64 bit issues with CGFloat
`CGFloat` can be a bit of a pain to work with since all the functions operating on floating points are bound to `float`, `double` and variants of those. To make it better I’ve introduced a set of `HUBCGFloat` prefixed functions. These functions should make it quite a bit nicer to work with `CGFloat` while staying true to the underlying type.

Instead of type casting the functions makes sure to use the correct function as provided by the OS. Depending on whether `CGFloat` is defined as a `double` or `float`.

Adds the following functions:
- `CGFloat HUBCGFloatFloor(CGFloat)` – returns the value, rounded down.
- `CGFloat HUBCGFloatAbs(CGFloat)` – returns the absolute value.
- `CGFloat HUBCGFloatMax(CGFloat, CGFloat)` – return the maximum value.
- `CGFloat HUBCGFloatMin(CGFloat, CGFloat)` – return the minimum value.
- `BOOL HUBCGFloatFloor(CGFloat, CGFloat, CGFloat)` – tests whether two values are approximately equal (±epsilon). Caveat: it doesn’t work if one of the values is `0.0`.
- `BOOL HUBCGFloatIsZero(CGFloat)` – tests whether the given value is approximately zero (±epsilon).
  - Should be used instead of `someFloatingPoint == 0.0` since that will likely yield a different result than expected. It also triggers a new build warning in Xcode 8.3.

### Additionally
Also fixes a newly introduced warning about comparing floating point values against 0. Given that floating point values and equality checks might yield in unexpected results. Such as:

```
10.0 == 10.0 + DBL_EPSILON  [yields true]
```

Xcode now warns about this:
screen shot 2017-02-20 at 17 53 07

The PR changes all current usage of the “normal” math variants of these functions have been updated. There are still some places where casting to `CGFloat` is necessary for the code to build for both 32 and 64 bit architectures.

Furthermore it fixes the type (and name) of the test helper macro to specify that it works on `CGFloats`.

Also removes the ignored warnings. Only the tests targets need to ignore partial availability (due to Apple’s XCTest bug). We didn’t catch the issue run into with Xcode 8.3 due to it being ignored for Xcode 7.3 -> 8.2. This is no longer the case.

### Review
I’ve tested the code and ran the tests using the iPhone 7 (64 bit, iOS 10.3) and iPhone 4s (32 bit, iOS 9.3) simulators.

For review by @spotify/objc-dev please 😸

### DIFF
--- a/sources/CGFloat+HUBMath.h
+++ b/sources/CGFloat+HUBMath.h
@@ -47,8 +47,8 @@ static inline CGFloat HUBCGFloatFloor(CGFloat value)
 /**
  Returns the absolute value of the given floating `value`.
  
- - `HUBFabs(±0)` returns `0`
- - `HUBFabs(±infinity) returns `+infinity`.
+ - `HUBCGFloatAbs(±0)` returns `0`
+ - `HUBCGFloatAbs(±infinity) returns `+infinity`.
 
  @param value The value.
  @return The absolute value.
@@ -113,6 +113,8 @@ static const CGFloat HUBCGFloatDefaultEpsilon = (CGFloat)0.00001;
  @param rhs The second value.
  @param epsilon The epsilon (i.e. tolerated deviance from true equality).
  @return `YES` if the `lhs` is approximately equal to `rhs` (±epsilon); otherwise `NO`.
+ 
+ @seealso `HUBCGFloatIsZero(CGFloat)` for a function to check if a values is ≈0.
  */
 static inline BOOL HUBCGFloatIsNearlyEqual(CGFloat lhs, CGFloat rhs, CGFloat epsilon)
 {
--- a/sources/CGFloat+HUBMath.h
+++ b/sources/CGFloat+HUBMath.h
@@ -1,0 +1,137 @@
+/*
+ *  Copyright (c) 2016 Spotify AB.
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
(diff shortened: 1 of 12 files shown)


marmelroy:
Strange that it's only causing a bug with Xcode 8.3 beta 2

rastersize:
@marmelroy Yeah, quite weird to be honest ¯\_(ツ)_/¯

sources/CGFloat+HUBMath.h
```diff
@@ -0,0 +1,139 @@
...
+ @param value The value that should be floored.
+ @return The largest integral value not greater than `value`.
+ */
+static inline CGFloat HUBCGFloatFloor(CGFloat value)
+{
+#if CGFLOAT_IS_DOUBLE
+    return floor(value);
+#else
+    return floorf(value);
+#endif // CGFLOAT_IS_DOUBLE
```
dflems: Can we use `tgmath.h` or C11 generic macros instead of our own helper functions?

rastersize: @dflems `tgmath.h` is still [broken with clang modules](http://www.openradar.me/15638008) AFAIK.

([SO topic](http://stackoverflow.com/questions/23333287/tgmath-h-doesnt-work-if-modules-are-enabled))

dflems: Noted! Either way, I think many of these functions can be simplified:

```objc
#if CGFLOAT_IS_DOUBLE
#define HUBCGFloatFloor floor
#define HUBCGFloatAbs fabs
#define HUBCGFloatMax fmax
#define HUBCGFloatMin fmin
#else
#define HUBCGFloatFloor floorf
#define HUBCGFloatAbs fabsf
#define HUBCGFloatMax fmaxf
#define HUBCGFloatMin fminf
#endif
```

rastersize:
Closing this PR in favor of #317. Since #317 takes this work and builds on it.

PR #289, opened 2017-02-20
Potential fix for ModuleMap bug in dynamic framework

marmelroy:
- Rename the module map so that it's not automatically discovered (matching target's name).
- Tell Xcode where it is.


marmelroy:
@rastersize mentioned potential issues with the S4A app and this kind of set up. Will run additional tests tomorrow. Don't merge yet.

marmelroy:
@rastersize is right, this is not working as well as I hoped.

PR #290, opened 2017-02-21
Multiple selection bug

bubski:
Up to this point you could select multiple cells at the same time using several fingers. This resulted in several new views being pushed.

This is fixed by making `HUBComponentGestureRecognizer`s not being allowed to handle touches more than one at a time.

cc: @mhallendal 👋

### DIFF
--- a/sources/HUBComponentGestureRecognizer.m
+++ b/sources/HUBComponentGestureRecognizer.m
@@ -52,13 +52,23 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)begin
 {
     self.state = UIGestureRecognizerStateBegan;
-    self.synchronizer.locked = YES;
+    [self.synchronizer gestureRecognizerDidBeginHandlingTouches:self];
+}
+
+- (void)beginIfPossible
+{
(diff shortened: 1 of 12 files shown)


sources/HUBComponentGestureRecognizer.m
```diff
@@ -20,55 +20,101 @@
...
     UITouch * const touch = [touches anyObject];
     CGPoint const touchLocation = [touch locationInView:self.view];
-    
-    if (touchLocation.y < 0 || touchLocation.y > CGRectGetHeight(self.view.bounds)) {
-        self.state = UIGestureRecognizerStateFailed;
-    } else if (touchLocation.x < 0 || touchLocation.x > CGRectGetWidth(self.view.bounds)) {
-        self.state = UIGestureRecognizerStateFailed;
+
+    if (!CGRectContainsPoint(self.view.bounds, touchLocation)) {
+        [self finishWithState:UIGestureRecognizerStateFailed];
```
mhallendal: Nice cleanup!

bubski:
Thanks.

Yes, that was my initial approach, but it felt a bit convoluted.
But I can reconsider and get back to you.

PR #291, opened 2017-02-22
HubViewModelRenderer: Add simple check for Header changes to diffing

quentar:
Account for changes of header component model.
Supplements PR https://github.com/spotify/HubFramework/pull/277 ( Don’t reload the collection view if there is nothing in the diff. )

Core of the change logic in `HubViewModelRenderer` :  "When header changes, do reload"
```objc  
BOOL const hasDiffChanges = (diff == nil || diff.hasBodyChanges || diff.hasHeaderChanges);
```

Added property `hasHeaderChanges` to HUBViewModelDiff class
Renamed property `hasChanges` to `hasBodyChanges` to be more specific
Added test cases for basic header changes
Expanded helper methods in  HUBViewModelUtilities.m to for creating header component easily

### DIFF
--- a/sources/HUBViewModelDiff.h
+++ b/sources/HUBViewModelDiff.h
@@ -70,8 +70,12 @@ extern HUBViewModelDiff *HUBDiffMyersAlgorithm(id<HUBViewModel>, id<HUBViewModel
 /// The index paths of any body components that were modified in the new view model. 
 @property (nonatomic, strong, readonly) NSArray<NSIndexPath *> *reloadedBodyComponentIndexPaths;
 
-/// A convenience property that returns YES if there are any inserts, deletes or reloads in this diff.
-@property (nonatomic, readonly) BOOL hasChanges;
+/// A convenience property that returns YES if there are any inserts, deletes or reloads in body components of this diff.
+@property (nonatomic, readonly) BOOL hasBodyChanges;
+
+/// A convenience property that returns YES if there is change in header component of this diff.
+@property (nonatomic, readonly) BOOL hasHeaderChanges;
+
 
 /**
  * Initializes a @c HUBViewModelDiff using the two view models by finding the longest common subsequence
--- a/sources/HUBViewModelDiff.h
+++ b/sources/HUBViewModelDiff.h
@@ -70,12 +70,8 @@ extern HUBViewModelDiff *HUBDiffMyersAlgorithm(id<HUBViewModel>, id<HUBViewModel
 /// The index paths of any body components that were modified in the new view model. 
 @property (nonatomic, strong, readonly) NSArray<NSIndexPath *> *reloadedBodyComponentIndexPaths;
 
-/// A convenience property that returns YES if there are any inserts, deletes or reloads in body components of this diff.
-@property (nonatomic, readonly) BOOL hasBodyChanges;
-
-/// A convenience property that returns YES if there is change in header component of this diff.
-@property (nonatomic, readonly) BOOL hasHeaderChanges;
-
+/// A convenience property that returns YES if there are any inserts, deletes or reloads in body or header components of this diff.
+@property (nonatomic, readonly) BOOL hasChanges;
 
 /**
  * Initializes a @c HUBViewModelDiff using the two view models by finding the longest common subsequence
--- a/sources/HUBViewModelDiff.m
+++ b/sources/HUBViewModelDiff.m
@@ -75,16 +75,12 @@ static inline NSArray<NSIndexPath *> *HUBIndexSetToIndexPathArray(NSIndexSet *in
     return [self diffFromViewModel:fromViewModel toViewModel:toViewModel algorithm:HUBDiffMyersAlgorithm];
 }
 
-- (BOOL)hasBodyChanges
+- (BOOL)hasChanges
 {
     return self.insertedBodyComponentIndexPaths.count > 0
         || self.deletedBodyComponentIndexPaths.count > 0
-        || self.reloadedBodyComponentIndexPaths.count > 0;
-}
-
-- (BOOL)hasHeaderChanges
-{
-    return self.headerComponentIdentifierHasChanged;
+        || self.reloadedBodyComponentIndexPaths.count > 0
+        || self.headerComponentIdentifierHasChanged;
 }
 
 - (void)calculateHeaderChangesFromViewModel:(id<HUBViewModel>)fromViewModel toViewModel:(id<HUBViewModel>)toViewModel
--- a/sources/HUBViewModelDiff.m
+++ b/sources/HUBViewModelDiff.m
@@ -22,6 +22,7 @@
 
 #import "HUBViewModelDiff.h"
 #import "HUBComponentModel.h"
+#import "HUBIdentifier.h"
 
 #import <UIKit/UIKit.h>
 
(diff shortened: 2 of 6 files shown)


sources/HUBViewModelDiff.h
```diff
@@ -70,8 +70,12 @@ extern HUBViewModelDiff *HUBDiffMyersAlgorithm(id<HUBViewModel>, id<HUBViewModel
 /// The index paths of any body components that were modified in the new view model. 
 @property (nonatomic, strong, readonly) NSArray<NSIndexPath *> *reloadedBodyComponentIndexPaths;
 
-/// A convenience property that returns YES if there are any inserts, deletes or reloads in this diff.
-@property (nonatomic, readonly) BOOL hasChanges;
```
cerihughes: I don't think we need to change this API. We're only interested in whether there are changes - we don't really care where they come from.

sources/HUBViewModelDiff.m
```diff
@@ -22,6 +22,7 @@
 
 #import "HUBViewModelDiff.h"
 #import "HUBComponentModel.h"
+#import "HUBIdentifier.h"
```
Goos: Is this import used?

sources/HUBViewModelDiff.m
```diff
@@ -37,6 +38,12 @@
     return [indexPaths copy];
 }
 
+@interface  HUBViewModelDiff ()
+
+@property (nonatomic, assign) BOOL headerComponentIdentifierHasChanged;
```
Goos: Wouldn't it make more sense to make this a public, readonly property? Since the rest of the diff model is immutable, and it might be useful for the consumer of the diff to see whether or not the header has been modified?

sources/HUBViewModelDiff.m
```diff
@@ -70,7 +79,22 @@ - (BOOL)hasChanges
...
+{
+    id<HUBComponentModel> fromHeaderModel = fromViewModel.headerComponentModel;
+    id<HUBComponentModel> toHeaderModel = toViewModel.headerComponentModel;
+
+    if (!fromHeaderModel && !toHeaderModel) {
+        return;
+    }
+
+    if (![fromHeaderModel isEqual:toHeaderModel]) {
+        self.headerComponentIdentifierHasChanged = YES;
```
Goos: Shouldn't this technically be `headerComponentHasChanged`, as you're comparing the entire model, rather than just the identifier?

sources/HUBViewModelDiff.m
```diff
@@ -57,7 +64,9 @@ + (instancetype)diffFromViewModel:(id<HUBViewModel>)fromViewModel
                         algorithm:(HUBDiffAlgorithm)algorithm
 {
     NSParameterAssert(algorithm);
-    return algorithm(fromViewModel, toViewModel);
+    HUBViewModelDiff *diff =  algorithm(fromViewModel, toViewModel);
+    [diff calculateHeaderChangesFromViewModel:fromViewModel toViewModel:toViewModel];
+    return diff;
```
Goos: I think it would make more sense to add this as a parameter to the initializer, and making the two algorithms perform the necessary comparison.

sources/HUBViewModelDiff.m
cerihughes: I actually made the opposite comment. None of the code that uses the diff cares which part of it has actually changed, so I don't think we need to change the API of this class.

sources/HUBViewModelDiff.m
cerihughes: Isn't the header comparison totally separate from the diffing algorithm though? It's only possible to assign a single header component so why would we want every algorithm to make the same comparison? Maybe I've misunderstood what the diffing algorithm is doing...

Goos: As I see it, the role of the diffing is to find a graph of changes between two view models – in which the header component resides. 
Implementation-wise, the header component is used a bit different right now due to it not being handled by the UICollectionView, but iirc, we would probably like to change that in the future.

sources/HUBViewModelDiff.m
Goos: Couldn't the API be a tad confusing though if the model reports no deletions/inserts/deletions, but still claims to have changes?

sources/HUBViewModelDiff.m
quentar: Good catch, thanks, forgot to update the name.

sources/HUBViewModelDiff.m
quentar: Thanks for noticing, not anymore I think, I'll update it.

sources/HUBViewModelDiff.m
quentar: I don't have strong opinions if to have two public APIs or one. Please decide if it is needed in scope of this PR as it can be of course changed later.

sources/HUBViewModelDiff.m
quentar: This method is de facto used 'initializer'. Adding check for header changes seemed better than just reverting PR #277 which seems to be useful.

Goos: I'm not saying we should change the public convenience constructor, just to introduce the `headerComponentHasChanged` parameter into the internal `initWithInserts:deletes:reloads:` initializer. This way the diffing algorithms would have full responsibility of all the diffing logic, and the `HUBViewModelDiff` would remain a immutable plain old data type.

sources/HUBViewModelDiff.m
Goos: I think it could be useful to expose it for the sake of clarity, but I'm fine either way. I'd rather it be `readonly` though 😛

sources/HUBViewModelDiff.m
cerihughes: But isn't the point of the diffing algorithm to tell us which components have been added, removed and changed? Isn't that not something that we can even do for the header component?
If we did delegate this responsibility to the algorithm, would any algorithm implement it in any other way than `[header1 equalTo:header2]`?

Goos: Yeah it would be implemented exactly the same way, only moved to a different place :smile:

I'm only arguing that since the prototype of the diffing function says _"Take these two view models and return me a complete list of changes between them"_, that should also include whether or not the header changed, as it is a part of the view model. If we modify the diff after the algorithm has run, then the algorithms no longer deliver a complete diff.

In addition to that, if we would ever want to change the header diffing, (e.g. by differentiating between a completely new header or in-place updates to an existing header), then this could differ between algorithms, as the header can have list of child components, much like the view model's body components.

cerihughes: What about when we include overlays in the diffing process (which we really should do). Wouldn't we want to reuse the diffing functions we already have for body components - i.e. remove the body specifics, make them more generic, then reuse them for the overlay components.

If we do that, considering the header components in those functions doesn't make sense either.

Goos: My point is that the diffing functions are defined as diffing view models, not body components. That is, one call to the diffing function should return the _entire_ diff, including body components, overlays and header.

We could (and probably should) change this if we wanted to diff multiple lists of components, but then it should return something else than a HUB_View_ModelDiff.

quentar:
!merge

PR #292, opened 2017-02-24
Trigger a load of images on rotation

cerihughes:
On rotation, we trigger a call to `[HUBComponentWrapper configureViewWithModel:containerViewSize:]`. This configures the cell to show the placeholder icon but doesn't reload the image.

Some components (e.g. the hugs carousel) will trigger a reload of images in all of its children at this point, but come components (e.g. the hugs entity card) don't. We should trigger a reload of components that support images at this point too.


cerihughes:
@spotify/objc-dev

PR #294, opened 2017-02-28
Expose bounce controls in HUBViewController

bubski:
```objc
@property (nonatomic, assign) BOOL bounces;
@property (nonatomic, assign) BOOL alwaysBounceVertical;
```
added to `HUBViewController` to allow controlling if the underlying `UICollectionView` should bounce when scrolled, as well as if it should allow scrolling vertically if content is smaller than the `UICollectionView`'s bounds.

bounce mov


bubski:
Good suggestion with the UITest.
Maybe it's even possible for me to check the `.conrtentOffset` of the `scrollView`?

I'll give it a shot

bubski:
@krris I've added unit and ui tests 🎾

PR #296, opened 2017-03-13
Delay observing of connectivity state until after view model should be loaded

albhaf:
If connectivity state updates before the viewmodel has been explicitly
loaded (i.e. its viewcontroller has appeared) then it might be loaded
unnecessarily at this point simply by being initialized.

Fixed by only observing connectivity after the point the view model
should be loaded.

### DIFF
--- a/sources/HUBViewModelLoaderImplementation.m
+++ b/sources/HUBViewModelLoaderImplementation.m
@@ -101,8 +101,6 @@ NS_ASSUME_NONNULL_BEGIN
         _cachedInitialViewModel = initialViewModel;
         _builderSnapshots = [NSMutableDictionary new];
         _errorSnapshots = [NSMutableDictionary new];
-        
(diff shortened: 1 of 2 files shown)


sources/HUBViewModelLoaderImplementation.m
```diff
@@ -174,6 +172,8 @@ - (BOOL)isLoading
 
 - (void)loadViewModel
 {
+    [self.connectivityStateResolver addObserver:self];
```
Goos: @albhaf Could there be any issues with calling `addObserver:` multiple times? As `loadViewModel` is called whenever the view controller appears?

Goos: Nevermind, it should be fine

albhaf: Yes, I checked that and it's idempotent

PR #298, opened 2017-03-13
Don’t track indexPathsByVerticalGroup

cerihughes:
The implementation of `layoutAttributesForElementsInRect:` produces several duplicates instances of `UICollectionViewLayoutAttributes`. Rewrote this implementation to just check frame intersections and in the tests that I ran, it produced exactly the same results but without duplication.

Removing this also allowed us to remove other private methods that are no longer needed.


cerihughes:
@spotify/objc-dev @mhallendal @Goos 

Thought I'd get this in before working on a larger refactoring of the layout.

PR #301, opened 2017-03-15
Remove coupling of HUBViewController and HUBViewModelLoaderImpl

mhallendal:
The HUBViewController was passed a concrete implementation of HUBViewModelLoader instead of the protocol which tightly couples it to the implementation.

This allows for a separate view model loader to be implemented (or passed in by the user of the framework).

I didn't want to expose the action performer on the HUBViewModel which is why the view controller checks for conformance to an internal subprotocol to HUBViewModelLoader.


mhallendal:
@spotify/objc-dev

PR #302, opened 2017-03-15
Live: Enable easier building & installation with SPM & Make

JohnSundell:
Hi everybody! 👋 Guess who's back? 😜

Played around a bit with Hubs & live editing again tonight, and realized we can make the command line tool for live editing a lot easier to install. This change makes `hublive` buildable using the Swift Package Manager, and installable through Make.

Instead of having users open and archive the Xcode project and manually move the built binary, `make` can now simply be run in the `live` folder, resulting in it being built using `swift build` and moved to `/usr/local/bin`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ documentation/output/
 
 # Ignore changes to our project.xcconfig that gets overridden by the build system
 project.xcconfig
+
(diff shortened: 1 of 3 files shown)


JohnSundell:
@Goos @rastersize @8W9aG 👋

PR #303, opened 2017-03-20
Refactor wrapper image loading and don't treat children differently.

cerihughes:
This PR refactors the component wrapper's image loading logic out of the `HUBViewController` and into a new `HUBComponentWrapperImageLoader` where it can be better understood and tested.

As part of this refactor, I noticed strange behaviour for loading of child images. There was some special logic gave a parent component the responsibility for loading images for child components when the child components were created (even though they might be offscreen at this point). The child components would then also be responsible for loading their own images once they appear on-screen.

There was also a bug in the logic that loaded the images when a child component was 1st created - when the child image loaded, the parent wrapper was informed and not the child so potentially, the image would appear in the wrong component. This was working mostly by coincidence as this initial load would ofter return early as the newly created child component was zero-sized at this point.

The PR simplifies the loading logic - now each component is responsible for loading its own images. Parents don't try and load images when children are created.

I've also moved as many tests as possible out of the view controller tests and into the image loader tests.


cerihughes:
@spotify/objc-dev @mhallendal

marmelroy:
+1

PR #306, opened 2017-03-22
Make hub manager optional

mhallendal:
As part of our work towards 3.x, we intend to simplify the framework by reducing the number of concepts and migrating parts which are more application specific out of the framework itself.  This includes aligning concepts with our Android version of Hubs (not yet publicly released) and reducing complexity.

As a first step we will be looking into the HUBFeatureRegistry and HUBManager. They are hard concepts to explain to new users of the framework, add a level of complexity and make the framework less flexible.

This PR adds a `HUBConfig` to take the role of the singleton `HUBManager` and will obsolete `HUBManager`, `HUBJSONRegistry`, `HUBFeatureRegistry`and the need to setup a global manager and register features to it.

The `HUBConfig` setup doesn't support the `HUBInitialViewModelRegistry` or Pre-/Append Content Operations. The current intention is to not support these.

List of things that needs to be fixed before this can be merged:

- [ ] Add support for `HUBConfig` with `HUBLiveService`.
- [ ] Add tests for the new code paths
- [ ] Add a comment about `HUBConfig` being created by `HUBConfigBuilder`.


demo/sources/AppDelegate.swift
```diff
@@ -71,7 +75,7 @@ import HubFramework
     }
     
     func applicationDidEnterBackground(_ application: UIApplication) {
-        hubManager.liveService?.stop()
+//        hubManager.liveService?.stop()
```
cerihughes: Let's get this working :)

include/HubFramework/HUBConfigViewControllerFactory.h
```diff
@@ -0,0 +1,60 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ *  `HUBViewController` factory that uses a `HUBConfig` for all the configuration needed to create a view controller.
+ *
+ *  This view controller allows for a fully configured `HUBViewController` to be created without going through the 
+ *  `HUBManager` and `HUBFeatureRegistration`.
+ */
+@interface HUBConfigViewControllerFactory : NSObject
```
cerihughes: I was expecting this to conform to `HUBViewControllerFactory`.

rastersize:
@mhallendal “Feature” request for 2.0 -> 3.0 and this PR: add a migration guide. That explains how to migrate existing code using the now removed or deprecated concept to the new world 🙇

include/HubFramework/HUBConfig.h
```diff
@@ -0,0 +1,91 @@
...
+@protocol HUBViewControllerScrollHandler;
+@protocol HUBComponentRegistry;
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ *  Configuration used to create view controllers when using Hub Framework without `HUBManager`.
+ *
+ *  This allows for using Hub Framework without the `HubManager` and feature registration.
```
rastersize: Could we add a sentence pointing the user to `HUBConfigBuilder`? The first questions that popped into my heads was “how am I supposed to create one of these objects? all the properties are `readonly`?”.

rastersize:
@mhallendal Also, with the removal of `HUBInitialViewModelRegistry` will there still be a way to add content to the page immediately? Before any, e.g., network loads?

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -353,7 +353,7 @@
 		8AE6C0A51DF6E40D0063B2B1 /* HUBComponentDefaults.m in Sources */ = {isa = PBXBuildFile; fileRef = 8A5D7A611CBBAC2200B987BA /* HUBComponentDefaults.m */; };
 		8AE6C0A61DF6E40D0063B2B1 /* HUBDefaultComponentLayoutManager.h in Headers */ = {isa = PBXBuildFile; fileRef = 8AA124E51DE89B530076582D /* HUBDefaultComponentLayoutManager.h */; };
 		8AE6C0A71DF6E40D0063B2B1 /* HUBDefaultComponentLayoutManager.m in Sources */ = {isa = PBXBuildFile; fileRef = 8AA124E61DE89B530076582D /* HUBDefaultComponentLayoutManager.m */; };
-		8AE6C0A81DF6E40D0063B2B1 /* HUBDefaultComponentFallbackHandler.h in Headers */ = {isa = PBXBuildFile; fileRef = 8AA127EC1DE8A1010076582D /* HUBDefaultComponentFallbackHandler.h */; };
+		8AE6C0A81DF6E40D0063B2B1 /* HUBDefaultComponentFallbackHandler.h in Headers */ = {isa = PBXBuildFile; fileRef = 8AA127EC1DE8A1010076582D /* HUBDefaultComponentFallbackHandler.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
mhallendal: It was, in order to expose the simple block based component fallback handler, however I'm not sure it makes sense. Especially if we are discussing removing the fallback handling.

include/HubFramework/HUBConfigViewControllerFactory.h
mhallendal: Initially it didn't make a lot of sense and later I didn't revisit that decision. It could be changed to be configured with a configuration (rather than passing in the configuration to the `create...`-methos). However at least one of the methods in the `HUBViewControllerFactory` protocol does't make sense when used without `HUBManager`/`HUBFeatureRegistry`.

mhallendal:
@rastersize Definitely agree about migration guide from 2.0 -> 3.0. I didn't add one for this step since I think/hope that we have been able to separate the various parts enough to obsolete the `HUBConfig` as well before we stabilise for 3.x.

mhallendal:
@rastersize re. initial view model registry. You can already handle it through the normal content loading by just having a content operation provide the initial view model without any network traffic.

Longer term we want to reverse the content loading (ie. the view controller has a setter for the view model and it's being fed from the outside, rather than pulled from the view controller. And in fact the view controller doesn't know about the view model loaders existens). That way the view model loader doesn't need to be involved in setting an initial view model.

include/HubFramework/HUBConfig.h
mhallendal: Definitely, I thought I did but apparently not. Thanks for spotting.

include/HubFramework/HUBConfigViewControllerFactory.h
cerihughes: Makes sense. I guess that means we're deprecating `HUBViewControllerFactory`, right?

PR #307, opened 2017-03-22
Allow different implementations of the HUBViewController

cerihughes:
This is the first step to us allowing the `HUBViewControllerFactory` to return different implementations of `HUBViewController`.

`HUBViewController` has essentially become an "abstract" class with no implementation. The "real" implementation has moved to `HUBViewControllerImplementation` which is a renamed version of the "old" `HUBViewController`.

We considered making `HUBViewController` a `@protocol` again, but felt that it would be too much of a change. It would also be a breaking API change which we're trying to avoid. There are also some categories in our internal Spotify use which would require rework.


cerihughes:
@mhallendal

cerihughes:
@spotify/objc-dev We're planning on eventually moving away from the existing `HUBViewController` (now renamed to `HUBViewControllerImplementation`) and towards a new implementation written from the ground up.

We'll soon support dynamic instantiation of different subclasses of `HUBViewController` based on some `HUBConfig` (which is coming in a later PR).

marmelroy:
+1

marmelroy:
I like the sound of HUBViewController as a @protocol again. Something to consider for 3.0.0?

cerihughes:
@marmelroy : There are arguments for and against it as a `protocol`, but we're definitely thinking about it.

PR #308, opened 2017-03-23
Allow users to register with options.

cerihughes:
This PR allows users of the HubFramework to register their features using a method that also exposes a `NSDictionary` of options.

The only use for this options dictionary (at present, although not currently implemented) is to allow Hubs to internally switch implementations of the `HUBViewController` on a per-feature basis.

To achieve this, the options dictionary is held in the `HUBFeatureRegistration` object which is available to the view controller factory.

*NOTE:* This will be a very short lived API which is for internal Spotify use only, so only use it if absolutely necessary. The concept will be available in future versions of Hubs (with the same caveats) but the API to this will be changing very soon as part of @mhallendal 's upcoming PR.

### DIFF
--- a/sources/HUBFeatureRegistration.m
+++ b/sources/HUBFeatureRegistration.m
@@ -33,6 +33,7 @@ NS_ASSUME_NONNULL_BEGIN
                customJSONSchemaIdentifier:(nullable NSString *)customJSONSchemaIdentifier
                             actionHandler:(nullable id<HUBActionHandler>)actionHandler
               viewControllerScrollHandler:(nullable id<HUBViewControllerScrollHandler>)viewControllerScrollHandler
+                                  options:(nullable NSDictionary<NSString *,NSString *> *)options
 {
     NSParameterAssert(featureIdentifier != nil);
     NSParameterAssert(featureTitle != nil);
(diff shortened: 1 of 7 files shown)


cerihughes:
@bubski @mhallendal @spotify/objc-dev

cerihughes:
@mhallendal This will probably conflict with the PR you're working on, but it should be pretty easy to carry the same concept over to your `HUBConfig` changes. It should be fine to have this extra API here for a short time... I've caveated its use in the documentation.

sources/HUBFeatureRegistration.m
```diff
@@ -50,6 +51,7 @@ - (instancetype)initWithFeatureIdentifier:(NSString *)featureIdentifier
         _customJSONSchemaIdentifier = [customJSONSchemaIdentifier copy];
         _actionHandler = actionHandler;
         _viewControllerScrollHandler = viewControllerScrollHandler;
+        _options = options;
```
bubski: Shouldn't this be `_options = [options copy]`?

ynmehel: +1

cerihughes: Agreed. I'll create a PR to fix it.

PR #309, opened 2017-03-24
Create experimental view controller implementation

cerihughes:
Create an "experimental" (and untested) view controller, and make it available if a feature registers with a `HUBViewController=v2` option.

The experimental view controller is a work-in-progress and is currently an exact line-for-line copy of the "standard" VC.


cerihughes:
@mhallendal

PR #310, opened 2017-03-24
Scroll Handler: add did scroll method

marmelroy:
We need to observe content offset changes inside a hub view controller and it currently doesn't implement a didScroll method. This PR adds one.

### DIFF
--- a/include/HubFramework/HUBViewControllerScrollHandler.h
+++ b/include/HubFramework/HUBViewControllerScrollHandler.h
@@ -99,6 +99,15 @@
                         currentContentRect:(CGRect)currentContentRect;
 
 /**
+ *  React to that an ongoing scrolling event in a view controller
+ *
+ *  @param viewController The view controller in question
+ *  @param contentOffest The scroll view's content offset
+ */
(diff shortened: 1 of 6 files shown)


krris:
@marmelroy Have you seen `HUBComponentContentOffsetObserver`? Maybe it will solve your problem without updating HubFramework. https://github.com/spotify/HubFramework/blob/master/include/HubFramework/HUBComponentContentOffsetObserver.h#L42

marmelroy:
@krris, I did. HUBComponentContentOffsetObserver reports the content offset to components so that they can update but I want to observe the contentOffset externally.

marmelroy:
Small issue: adding a new method in the scroll handler protocol is technically a breaking change as it requires updating all classes that conform to the protocol.

I did consider making it an optional method but the fact that the scroll handler is of an "id" type and doesn't inherit from NSObject means I can't check if the the delegate responds to the optional selector.

cerihughes:
If you could create a github issue that addresses the duplication of behaviour that this adds, we can go back to it later and work on that.

PR #313, opened 2017-03-27
Update experimental VC to not use batch updates

cerihughes:
We will further develop the experimental VC (or other experimental VCs) to be more modular.


cerihughes:
@mhallendal

ynmehel:
👍

PR #315, opened 2017-03-29
Copy the parameter as it has a copy attribute on the property.

cerihughes:
Fixes some code that was introduced in https://github.com/spotify/HubFramework/pull/308 and caught in review after the merge.


cerihughes:
@bubski @ynmehel FYI

ynmehel:
👍

PR #317, opened 2017-03-30
Fix building static lib with Xcode 8.3

cerihughes:
I took @rastersize 's PR (https://github.com/spotify/HubFramework/pull/288) and made the following additions:
- resolved merge conflicts,
- applied @dflems ' suggestion of using #define for some math operations
- added a workaround for a known problem with Xcode 8.3 and modulemaps (see "Known Issues" under "Apple LLVM Compiler" at https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW160)

Note: Some UI tests are still failing for me locally in Xcode8.3 - I'm submitting this PR to see if the build system (using a previous version of Xcode) has the same problem.


cerihughes:
Confirmed: Our ui test failures are Xcode8.3-specific.

HubFramework.xcodeproj/project.pbxproj
```diff
@@ -2222,31 +2232,27 @@
...
-				HUB_NON_ERROR_WARNINGS_0720 = "";
-				HUB_NON_ERROR_WARNINGS_0730 = "$(HUB_NON_ERROR_WARNING_0720) -Wno-error=partial-availability";
 				INFOPLIST_FILE = tests/Info.plist;
 				PRODUCT_BUNDLE_IDENTIFIER = com.spotify.HubFrameworkTests;
 				PRODUCT_NAME = HubFrameworkTests;
 				WARNING_CFLAGS = (
 					"$(inherited)",
-					"$(HUB_NON_ERROR_WARNINGS_$(XCODE_VERSION_MINOR))",
 					"-Wno-documentation-unknown-command",
+					"-Wno-partial-availability",
```
dflems: This is for test targets when you import `XCTest` right? I'm pretty sure you can do this:

```objc
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wpartial-availability"
#import <XCTest/XCTest.h>
#pragma clang diagnostic push
```

I think you can just slap that in its own header like `HUBXCTest.h` then just import that header at the top of your test classes.

dflems: @rastersize?

cerihughes: @dflems : Yes - as I understand it, we get partial availability errors when we import `XCTest`. Having another header to import just to get around this feels a bit like a hacky workaround, but I agree - switching off all partial availability warnings is a bit scary. Any ideas @rastersize @JensAyton ?

JensAyton: Partial availability warnings in test targets don’t strike me as important enough to worry about.

cerihughes: Fair point. @dflems - are you happy to give the thumb?!

rastersize: I’m good with disabling partial availability warnings for the unit test target. We can always revisit that specific issue in a separate PR.

sources/HUBViewControllerDefaultScrollHandler.m
```diff
@@ -89,16 +91,16 @@ - (CGPoint)contentOffsetForDisplayingComponentAtIndex:(NSUInteger)componentIndex
...
     } else if (scrollPosition & HUBScrollPositionBottom) {
         targetOffset = CGRectGetMaxY(componentFrame) - viewHeight;
     } else {
         // Default to putting it at the top unless a proper position is provided
         targetOffset = CGRectGetMinY(componentFrame);
     }
 
-    targetOffset = MAX(-contentInset.top, MIN(contentSize.height - viewHeight, targetOffset));
-    return CGPointMake(0.0, (CGFloat)floor((double)targetOffset));
+    targetOffset = HUBCGFloatMax(-contentInset.top, MIN(contentSize.height - viewHeight, targetOffset));
```
rastersize: @cerihughes Should we use `HUBSCGFloatMin` here (the nested one)?

cerihughes: Yep. Good point.

cerihughes: https://github.com/spotify/HubFramework/pull/322

PR #318, opened 2017-03-30
Make HUBLiveService independent from HUBViewController

mhallendal:
Instead of the HUBLiveService creating a view controller, have it emit a content operation and let the implementor setup and push the view controller. It makes the live service independent of the view controller and leaves the user in charge of what should be done (for example, add more content operations to the chain.

**NOTE:** This introduces a small API break in that it changes the `HUBLiveServiceDelegate` protocol. Given that this change makes sense even without the changes I'm making to the feature registry I deemed it made sense to clean this up on master even if we'll need to update the client when bumping this.

@cerihughes @spotify/objc-dev @rastersize @mumme

### DIFF
--- a/sources/HUBLiveServiceImplementation.m
+++ b/sources/HUBLiveServiceImplementation.m
@@ -21,7 +21,6 @@
 
 #import "HUBLiveServiceImplementation.h"
 
-#import "HUBViewControllerFactory.h"
 #import "HUBLiveContentOperation.h"
 
 #if HUB_DEBUG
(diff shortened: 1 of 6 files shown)


sources/HUBLiveServiceImplementation.m
```diff
@@ -128,24 +113,17 @@ - (void)handleBytesAvailableForStream:(NSInputStream *)stream
...
-    
-    HUBViewController * const viewController = [self.viewControllerFactory createViewControllerForViewURI:viewURI
-                                                                                        contentOperations:@[contentOperation]
-                                                                                        featureIdentifier:@"live"
-                                                                                             featureTitle:@"Hub Framework Live"];
-    
-    self.viewController = viewController;
-    [self.delegate liveService:self didCreateViewController:viewController];
+
+    [self.delegate liveService:self didCreateContentOperation:contentOperation];
```
cerihughes: Does this mean we have to create a new VC every time some data comes in over the wire?

mhallendal: No, as long as we hold on to the content operation (which the view controller will do) it will be reused by the live service and the delegate method is not called.

cerihughes: 👍

cerihughes:
👍

mhallendal:
This was merged as part of #306.

PR #320, opened 2017-03-30
Rewrite swipe tests to not count swipes

cerihughes:
The behaviour of `XCUIElement`'s `swipeUp` and `swipeDown` methods seems to have changed with iOS10.3 (swipes now seem to stop when a cell is completely on screen and don't decelerate).

To make sure our swiping tests work between versions, I've rewritten them to swipe until the desired header height is reached. They'll then swipe once more to make sure the bounce happens, and the height settles back to the desired height after the bounce.

If the expected height hasn't been reached by 20 swipes, the test fails (as before).

_Note_: Can someone validate my Swift method naming - I'm still making the transition ;)

### DIFF
--- a/demo/tests/HeaderComponentUITests.swift
+++ b/demo/tests/HeaderComponentUITests.swift
@@ -30,8 +30,9 @@ class HeaderComponentUITests: UITestCase {
         XCTAssertEqual(header.frame.height, 250)
         
         let collectionView = XCUIApplication().collectionViews.element(boundBy: 0)
-        collectionView.swipeUp()
-        XCTAssertEqual(header.frame.height, 64)
+        execute(block: {
+            collectionView.swipeUp()
+        }, until: header, hasHeight: 64)
     }
     
     func testCollectionViewContentInsetEqualToHeaderHeight() {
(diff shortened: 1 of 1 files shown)


cerihughes:
@spotify/objc-dev

demo/tests/HeaderComponentUITests.swift
```diff
@@ -80,18 +81,38 @@ class HeaderComponentUITests: UITestCase {
...
+        }, until: header, hasHeight: 250)
     }
     
     // MARK: - Utilities
     
     private func navigateToStickyHeaderFeature() {
         XCUIApplication().collectionViews.staticTexts["Sticky header"].tap()
     }
+
+    private func execute(block:() -> Void, until header:XCUIElement, hasHeight expectedHeight:CGFloat) {
```
mhallendal: It's usually preferred if you can put the block as the last parameter. This will have two benefits:
1. The call site will be a lot easier to read with:
``` swift
execute(until: header, hasHeight: 250, block: ^{})
```

2. When the block is the last parameter you can even leave out that as an argument when calling:
``` swift
execute(until: header, hasHeight: 250) {
    collectionView.swipeDown()
}
```

PR #322, opened 2017-04-03
Change MIN to GLUECGFloatMin

cerihughes:
I missed this in the last review.


cerihughes:
1 last case of MIN that was missed in https://github.com/spotify/HubFramework/pull/317

cerihughes:
@rastersize

PR #323, opened 2017-04-04
Rename modulemap back to module.modulemap

mhallendal:
This solves the problem we have been having when compiling other targets
using the static library.

From what I could tell the issues with the module.modulemap file was due
to the dynamic framework target creating its own module map file when
modules are enabled.

I think changing this only makes @imports unavailable internally in the
framework but we don't use those anyway.

@cerihughes @spotify/objc-dev


cerihughes:
👍 Let's try it!

PR #327, opened 2017-04-07
Always create collection view in load view

cerihughes:
We removed the code that released the collection view on low memory. It makes sense to have a single point of creation for the view again as it simplifies the VC.

I've also removed the configuration of the `bounce` property. It's currently never set to anything other than the default. I'll work on a PR that exposes the view through the API, so clients can set things directly on the view in future - this will mean that we don't have to write so much "pass through" code.

### DIFF
--- a/sources/HUBViewControllerExperimentalImplementation.m
+++ b/sources/HUBViewControllerExperimentalImplementation.m
@@ -92,7 +92,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @implementation HUBViewControllerExperimentalImplementation
 
-@synthesize bounces = _bounces;
 @synthesize alwaysBounceVertical = _alwaysBounceVertical;
 @synthesize viewModel = _viewModel;
 @synthesize viewURI = _viewURI;
@@ -141,7 +140,6 @@ NS_ASSUME_NONNULL_BEGIN
     _componentWrappersByCellIdentifier = [NSMutableDictionary new];
     _componentWrappersByModelIdentifier = [NSMutableDictionary new];
     _renderingOperationQueue = [HUBOperationQueue new];
-    _bounces = YES;
 
     viewModelLoader.delegate = self;
     if (HUBConformsToProtocol(viewModelLoader, @protocol(HUBViewModelLoaderWithActions))) {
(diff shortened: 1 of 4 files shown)


cerihughes:
@mhallendal @bubski @ynmehel @krris @poncel

ynmehel:
👍

sources/HUBViewControllerExperimentalImplementation.m
```diff
@@ -163,9 +163,22 @@ - (void)dealloc
 
 - (void)loadView
 {
-    self.view = [[HUBContainerView alloc] initWithFrame:CGRectZero];
+    HUBContainerView *containerView = [[HUBContainerView alloc] initWithFrame:CGRectZero];
```
krris: You can add `const` to have a consistency across our codebase.

sources/HUBViewControllerImplementation.m
```diff
@@ -167,9 +167,22 @@ - (void)dealloc
 
 - (void)loadView
 {
-    self.view = [[HUBContainerView alloc] initWithFrame:CGRectZero];
+    HUBContainerView *containerView = [[HUBContainerView alloc] initWithFrame:CGRectZero];
```
krris: Same thing as before: `const` 😉

sources/HUBViewControllerImplementation.m
```diff
@@ -167,9 +165,23 @@ - (void)dealloc
...
 
-    [self createCollectionViewIfNeeded];
+    HUBCollectionView * const collectionView = [self.collectionViewFactory createCollectionView];
+    self.collectionView = collectionView;
+    collectionView.showsVerticalScrollIndicator = [self.scrollHandler shouldShowScrollIndicatorsInViewController:self];
+    collectionView.showsHorizontalScrollIndicator = collectionView.showsVerticalScrollIndicator;
+    collectionView.keyboardDismissMode = [self.scrollHandler keyboardDismissModeForViewController:self];
+    collectionView.decelerationRate = [self.scrollHandler scrollDecelerationRateForViewController:self];
+    collectionView.bounces = YES;
```
krris: Are you sure that none of the views in our app set it to `NO`?

cerihughes: The API isn't used by anything. `client-ios` uses `alwaysBounceVertical`, but I'll be moving that API into the view layer.

sources/HUBViewControllerImplementation.m
cerihughes: Oh. Remembering to use `const` isn't one of my strong points :O

PR #329, opened 2017-04-11
Use dependency injection for UIApplication.

krris:
Currently, `UIApplication` is accessed by `[UIApplication sharedApplication]` in many places. It makes it hard to test. Especially, when we want to mock `UIApplication`, which can be used by other frameworks.

This PR introduces `HUBApplicationProtocol` and `HUBApplication` (which is a wrapper around `UIApplication`).

🎾 @cerihughes @mhallendal @poncel @ynmehel


sources/HUBActionRegistryImplementation.h
```diff
@@ -36,8 +37,10 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  To be able to specify which selection action to use (useful for tests), use this class'
  *  designated initializer instead of this class constructor.
+ *
+ *  @param application The object exposing UIApplication's properties and methods
  */
-+ (instancetype)registryWithDefaultSelectionAction;
++ (instancetype)registryWithDefaultSelectionActionAndApplication:(id<HUBApplicationProtocol>)application;
```
krris: I’m not happy about naming here. Do you have any better ideas?

cerihughes:
I don't think we need to inject the `UIApplication` any further up (or is it down?) the stack than the VC... I think it's OK for the "root" objects like `HUBConfigBuilder` and `HUBManager` to call `[UIApplication sharedApplication]` internally and inject that into everything else that needs it downstream. Maybe there's a more suitable "root" object in Hubs that we could use (but I think these 2 are about as "root" as it gets).

This is much like `client-ios` where root objects like service/feature implementations generally access static singletons and inject them into all objects from that point (or at least, that's how *I* think it should be working).

I don't think we should be exposing the application through the API as a `HUBApplication` either as that's confusing for users. Doing the above means we don't have to do this either

cerihughes:
... in other words, I don't think there should be any changes to the `include` APIs for this change. It should be purely an implementation change.

krris:
@ceri I changed a little the implementation. Now the root objects create and inject HUBApplication. Unfortunately, we still need to expose HUBApplication header to be able to mock it.

cerihughes:
@krris : I'm struggling to see how injecting a `HUBApplication` is any easier than injecting a `UIApplication`. Is there something that doesn't work if we just use `UIApplication`?

cerihughes:
Oh... is it because we don't have a mocking framework and so can't properly mock a `UIApplication`?

krris:
@cerihughes Yes, it’s because we don’t have a mocking framework.

cerihughes:
@krris Can we just subclass `UIApplication` for testing purposes (in the same way we do for `HUBCollectionViewMock` for example)?

krris:
@cerihughes Unfortunately we cannot. If we create `HUBApplicationMock` which is a subclass of `UIApplication`, we won't be able to instantiate it, because there can only be one `UIApplication` instance.

cerihughes:
We *do* have the option of subclassing `NSObject` instead of `UIApplication` and just mirroring the API that we use on that subclass.

I'm all for writing code that's easy to test, but I think that needing changes to the public API just to be testable is a step too far. WDYT?

PR #330, opened 2017-04-12
HubViewController as protocol

marmelroy:
@cerihughes and I discussed the fact that subclassing HubViewController in order to define a contract (i.e. act as a protocol) was a bit strange. 

The argument against it was that Swift will look at UIViewController<HubViewController> as just a UIViewController and there will be lots of casting in the Swift code. 

That's true (for now).

However,  Swift looks at id<HubViewController> as a HubViewController protocol. 

Going with this approach, there's only one line of casting in the Swift code and that's only to make UIKit happy. I think that's acceptable.


cerihughes:
... but if we *do* create a protocol for the VC (there are still other cases where it's easier if it's a concrete class), we'd want to expose this in the API as `UIViewController<HUBViewController> *`, not `id<HUBViewController>`.

I think we should get to a point where we're not experimenting so much with different VC implementations and then look at whether using a protocol in the API is feasible.

For now, let's keep it as it is.

rastersize:
@marmelroy @cerihughes With Swift 4 it’ll be possible to expose and use types as `UIViewController<HUBViewController>`. So “all” will be good in about 6 months time.

marmelroy:
@rastersize, can't wait :)

PR #332, opened 2017-04-21
Fix duplicate selection logic

rastersize:
The logic to stop multiple gesture recognizers where a bit flawed. When a recognizer was told to not recognize it would immediately reset the state of the synchronizer. Such that the next recognizer would be able to recognize. Even if the original recognizer hadn’t stopped recognizing. We now compare the finished gesture recognizer with the one that began handling the gesture. Only if it’s the same do we reset it; allowing another one to recognize.

Both the collection view and our custom selection logic was firing. This could cause a component to stay selected visually until the view was interacted with again. As such `allowsSelection` has been set to `NO` allowing HubFramework to do its custom selection and highlighting management.

@spotify/objc-dev

### DIFF
--- a/sources/HUBSingleGestureRecognizerSynchronizer.m
+++ b/sources/HUBSingleGestureRecognizerSynchronizer.m
@@ -22,24 +22,28 @@
 #import "HUBSingleGestureRecognizerSynchronizer.h"
 
 @interface HUBSingleGestureRecognizerSynchronizer ()
-@property (nonatomic, assign) BOOL shouldPreventGestureRecognizersFromHandlingTouches;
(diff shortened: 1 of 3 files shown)


mhallendal:
Looks reasonable to me. I stumbled over the fact that we both allow selection in the `UICollectionView` and have a gesture recognizer for it as well. Just waiting for @bubski to have a chance to give feedback to make sure this will work for his use case as well.

bubski:
@rastersize @mhallendal Yes, that makes sense. I don't see why I didn't do it like that. Go ahead

bubski:
+1

PR #335, opened 2017-05-02
Bug: wrong argument is sent to HUBCollectionViewDelegate method

krris:
The wrong argument is sent to the following HUBCollectionViewDelegate method:
```
- (void)collectionView:(UICollectionView *)collectionView
  didEndDisplayingCell:(UICollectionViewCell *)cell
    forItemAtIndexPath:(NSIndexPath *)indexPath
```

Instead of `HUBComponentCollectionViewCell` we should send a view from the component wrapper.

### DIFF
--- a/sources/HUBViewControllerExperimentalImplementation.m
+++ b/sources/HUBViewControllerExperimentalImplementation.m
@@ -717,6 +717,7 @@ willUpdateSelectionState:(HUBComponentSelectionState)selectionState
 {
     HUBComponentWrapper * const componentWrapper = [self componentWrapperFromCell:(HUBComponentCollectionViewCell *)cell];
     UIView * const componentView = componentWrapper.view;
+    NSAssert(componentView != nil, @"Component view shouldn't be nil");
 
     if (componentView) {
         [self.delegate viewController:self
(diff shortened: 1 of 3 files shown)


sources/HUBViewControllerExperimentalImplementation.m
```diff
@@ -716,11 +716,14 @@ - (void)collectionView:(UICollectionView *)collectionView
     forItemAtIndexPath:(NSIndexPath *)indexPath
 {
     HUBComponentWrapper * const componentWrapper = [self componentWrapperFromCell:(HUBComponentCollectionViewCell *)cell];
+    UIView * const componentView = componentWrapper.view;
 
-    [self.delegate viewController:self
-               componentWithModel:componentWrapper.model
-                     layoutTraits:componentWrapper.layoutTraits
-             didDisappearFromView:cell];
+    if (componentView) {
```
cerihughes: If we're not expecting the view to be `nil`, maybe we should add an assert here?

krris:
@cerihughes 🎾

sources/HUBViewControllerImplementation.m
```diff
@@ -727,11 +727,14 @@ - (void)collectionView:(UICollectionView *)collectionView
     forItemAtIndexPath:(NSIndexPath *)indexPath
 {
     HUBComponentWrapper * const componentWrapper = [self componentWrapperFromCell:(HUBComponentCollectionViewCell *)cell];
+    UIView * const componentView = componentWrapper.view;
```
cerihughes: Also here?

PR #352, opened 2017-10-03
Set accessibility identifier by default

rastersize:
Will be set to the component instance’s identifier. To make it easier to find Hub components in UI tests.

@spotify/objc-dev @mumme


mumme:
👍

mumme:
👍

PR #356, opened 2017-11-21
Fix non-portable include path

dflems:
Xcode seems to be ok with non-portable includes when using headermaps (the xcode default behavior). We have some other tooling that builds without header maps and this code generates a warning because of the misspelling.

@rastersize @cerihughes @8W9aG


8W9aG:
👍

PR #357, opened 2017-11-21
Update Gemfile & Fix Codecov

dflems:
Update `Gemfile` to silence a warning. Also, pass `-X xcodeplist` to codecov uploader to scan for and process `xccoverage` reports instead of using `llvm-cov`.


8W9aG:
👍

PR #366, opened 2018-06-07
Xcode 10b1 fixes

rastersize:
This makes the project build with Xcode 10b1 without analyzer warnings.

I’ve opted to make the call crash (via an assert) if we can’t allocate more memory. To make sure we don’t touch arbitrary memory I’ve also made sure asserts are enabled in RELEASE builds.

@dflems @8W9aG


rastersize:
I had to do some more changes to make it test OK:
- Upgrade to using Xcode 9.3 on Travis-CI.
  - They don’t support Xcode 9.4 yet.
- Upgrade demo app and live utility to Swift 4.
- Add support for running tests on iOS 11.
  - One test has been disabled.
  - Some tests had to be modified to be able to find the navigation bar title.
  - Disable iOS’ automatic content insets adjustment as HubFramework does not work well with it.

rastersize:
This PR should also make `master` green again 😄

